PR #23, opened 2019-03-21
Load WebView separately

andresilveirah:
TL;DR; We now: 
* load the webview in a separate function and call `onMessageReady` when the message is ready to be shown. 
* no longer add/remove the view from the superview. it's up to the parent to decide if/when the view should be added (we recommend using `onMessageReady` to add it and `onInteractionComplete` to remove it)

### DIFF
--- a/ConsentViewController/Classes/ConsentViewController.swift
+++ b/ConsentViewController/Classes/ConsentViewController.swift
@@ -406,9 +406,66 @@ import Reachability
         }
         userDefaults.setValue(String(parsedPurposeConsents), forKey: ConsentViewController.IAB_CONSENT_PARSED_PURPOSE_CONSENTS)
     }
+    
+    private func onReceiveMessage(willShow: Bool) {
+        if(willShow) {
+            webView.frame = webView.superview!.bounds
+            willShowMessage?(self)
+        } else { done() }
+    }
 
-    private func isDefined(_ string: String) -> Bool {
-        return string != "undefined"
+    private func onMessageChoiceSelect(choiceType: Int) {
+        guard Reachability()!.connection != .none else {
+            onErrorOccurred?(NoInternetConnection())
+            done()
+            return
+        }
+        self.choiceType = choiceType
+        onMessageChoiceSelect?(self)
+    }
+
+    private func onInteractionComplete(euconsent: String, consentUUID: String) {
+        self.euconsent = euconsent
+        self.consentUUID = consentUUID
+        do {
+            try storeIABVars(euconsent)
+            let userDefaults = UserDefaults.standard
+            userDefaults.setValue(euconsent, forKey: ConsentViewController.EU_CONSENT_KEY)
+            userDefaults.setValue(consentUUID, forKey: ConsentViewController.CONSENT_UUID_KEY)
+            userDefaults.synchronize()
+        } catch let error as ConsentViewControllerError {
+            onErrorOccurred?(error)
+        } catch {}
+        done()
+    }
+
+    private func onErrorOccurred(errorType: String) {
+        onErrorOccurred?(WebViewErrors[errorType] ?? PrivacyManagerUnknownError())
+        done()
+    }
+
+    private func handleMessage(withName name: String, andBody body: [String:Any?]) {
+        switch name {
+        case "onReceiveMessageData": // when the message is first loaded
+            onReceiveMessage(willShow: body["willShowMessage"] as! Int == 1)
+        case "onMessageChoiceSelect": // when a choice is selected
+            onMessageChoiceSelect(choiceType: body["choiceType"] as! Int)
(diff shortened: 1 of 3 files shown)


Example/ConsentViewController/ViewController.swift
```diff
@@ -10,84 +10,63 @@ import ConsentViewController
...
-                    cvc.euconsent as Any,
-                    "\n consent uuid prop",
-                    cvc.consentUUID as Any,
-                    "\n eu consent in storage",
-                    UserDefaults.standard.string(forKey: ConsentViewController.EU_CONSENT_KEY) as Any,
-                    "\n consent uuid in storage",
-                    UserDefaults.standard.string(forKey: ConsentViewController.CONSENT_UUID_KEY) as Any,
+            consentViewController.onMessageReady = { controller in
+                parentView.addSubview(controller.view)
+                controller.view.frame = controller.view.superview!.bounds
```
kavikumra: Could we remove this force unwrap by accessing the `parentView.bounds` directly? Also, might make it a bit clearer that the `parentView` size is being adhered to!

```suggestion
                controller.view.frame = parentView.bounds
```

ConsentViewController/Classes/ConsentViewController.swift
```diff
@@ -407,38 +416,47 @@ import Reachability
...
+    private func handleMessage(withName name: String, andBody body: [String:Any?]) {
         switch name {
         case "onReceiveMessageData": // when the message is first loaded
-            if(body["willShowMessage"] as! Int) == 1 {
-                webView.frame = webView.superview!.bounds
-                willShowMessage?(self)
-                return
-            }
-            done()
+            onReceiveMessage(willShow: body["willShowMessage"] as! Int == 1)
```
kavikumra: Is there any risk that these values could not be `Int` (if the JSON wasn't quite right, for example?)

If it could, it would be nice to have this as an optional check, and if it can't be parsed as an `Int` call the default behaviour for the switch.

ConsentViewController/Classes/ConsentViewController.swift
```diff
@@ -274,12 +275,20 @@ import Reachability
                 return
             }
             webView.load(URLRequest(url: messageUrl))
+            messageStatus = .loaded
         } catch let error as ConsentViewControllerError {
+            messageStatus = .notStarted
```
kavikumra: Do we need a similar step to this one, where you reset the `messageStatus` in the `Reachability` guard above? Otherwise if it errors due to network connection, we won't be able to retry!

andresilveirah: Good point. My idea, at first, wasn't to make it retriable, but simply to inform about the status of the message. But, in any case, we should reset `messageStatus` to `.notStarted` in the guard clause. 
Regarding  when `setSubjectToGDPR` throw an error, this will be captured by the `catch` and there we already reset the message status.

ConsentViewController/Classes/ConsentViewController.swift
andresilveirah: Since I treat SourcePoint's api as a black box, we should prepare for the case when the JSON isn't well formed. So yeah, I'll make it optional checks.

Example/ConsentViewController/ViewController.swift
andresilveirah: Yep. Much easier.

andresilveirah:
@kavikumra would you mind having another look?

ConsentViewController/Classes/ConsentViewController.swift
```diff
@@ -449,23 +449,21 @@ import Reachability
...
-                done()
-                return
-            }
+            else { fallthrough }
             onInteractionComplete(euconsent: euconsent, consentUUID: consentUUID)
         case "onErrorOccurred":
             onErrorOccurred(errorType: body["errorType"] as? String ?? "")
         default:
-            print("userContentController was called but the message body: \(name) is unknown.")
+            print("Couldn't parse message in userContentController. Called with name: \(name) and body: \(body)")
```
kavikumra: I don't think this case will ever be called now, as the `onErrorOccurred` case above doesn't have any guard - probably isn't a problem, but `onErrorOccurred` will be called if it's failed the other cases, and will contain no error message if it can't find `body["errorType"]`.

Are you okay with that, or do you want to ensure it can get through to `default`?

andresilveirah: Not really @kavikumra . If there's no `body["errorType"]` then it'd be evaluated to "" (as its default ?? value) and when looking up for that error type, we'd end up with `PrivacyManagerUnknownError`. Either way, I've added an extra error for the default case and when the message name and body can't be parsed (https://github.com/SourcePointUSA/ios-cmp-app/pull/23/files#diff-76246d3a966ad7ce8b0bd0f75d46a94eR477)

PR #31, opened 2019-04-30
Changed ReachabilitySwift version to 4.3.1 to provide support for Objective-C based application.

vilasma:
Changed ReachabilitySwift version to 4.3.1 to provide support for Objective-C based application.


andresilveirah:
@vilasma I believe we can close this PR right? As we're not updating Reachability (at least not until we update to Swift 5)

vilasma:
Yes, we can close this as we are not updated with swift 5.

PR #38, opened 2019-06-23
Implement auth

andresilveirah:
@vilasma when reviewing this PR, ignore the last commit. It's just an example app to show case the identity feature.

* implement `ConsentViewController.loadMessage(forAuthId: String)`
* update `euconsent` and `consentUUID` also when message is not displayed


vilasma:
Hi @andresilveirah I had gone through the code changes for auth ID except last commit i.e. auth example

PR #43, opened 2019-08-09
Replaced webview with new one and SDK flow is changed.

vilasma:
Replaced webview with new one and SDK flow is changed, now I am able to see message on iOS version 12.1.
@andresilveirah if you think changes for clients are minimum then we can think on optimization else we discard this option.

### DIFF
--- a/ConsentViewController/Classes/ConsentViewController.swift
+++ b/ConsentViewController/Classes/ConsentViewController.swift
@@ -10,8 +10,8 @@
 public typealias Callback = (ConsentViewController) -> Void
 
 import UIKit
-import WebKit
-import JavaScriptCore
+//import WebKit
+//import JavaScriptCore
 import Reachability
 
 /**
@@ -39,7 +39,7 @@ import Reachability
  view.addSubview(consentViewController.view)
  ```
 */
-@objcMembers open class ConsentViewController: UIViewController, WKUIDelegate, WKNavigationDelegate, WKScriptMessageHandler {
+@objcMembers open class ConsentViewController: UIViewController, ConsentWebViewHandler  {
 
     /// :nodoc:
     public enum DebugLevel: String {
@@ -122,8 +122,6 @@ import Reachability
 
     public var onErrorOccurred: ((ConsentViewControllerError) -> Void)?
 
-    var webView: WKWebView!
-
     /// Holds the choice type the user has chosen after interacting with the ConsentViewController
     public var choiceType: Int? = nil
 
@@ -143,6 +141,9 @@ import Reachability
     private let sourcePoint: SourcePointClient
 
     private var newPM = false
+    
+    public var sendURL: URL?
+    public var webview: WebView?
 
     /**
      Initialises the library with `accountId` and `siteName`.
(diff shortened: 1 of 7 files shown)


ConsentViewController/Classes/WebView.swift
```diff
@@ -0,0 +1,147 @@
...
+//  ConsentViewController
+//
+//  Created by Vilas on 8/7/19.
+//
+
+import UIKit
+import WebKit
+import JavaScriptCore
+
+public protocol ConsentWebViewHandler {
```
andresilveirah: What if we define the protocol here and make it mandatory for our client's ViewController to implement it? Then they'd have to add call something like:
```swift
consentViewController.consentDelegate(self)
```
What you think?

ConsentViewController/Classes/WebView.swift
```diff
@@ -0,0 +1,147 @@
...
+    func willShowMessage()
+    func didGetConsentData(euconsent: String, consentUUID: String)
+}
+
+@objcMembers open class WebView: UIView, UIWebViewDelegate , WKNavigationDelegate, WKScriptMessageHandler {
```
andresilveirah: What about giving the `WebView` a more descriptive name? Something like `ConsentWebView`.

ConsentViewController/Classes/ConsentViewController.swift
```diff
@@ -143,6 +141,9 @@ import Reachability
     private let sourcePoint: SourcePointClient
 
     private var newPM = false
+    
+    public var sendURL: URL?
```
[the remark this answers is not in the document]
vilasma: Best name suggested by you.

ConsentViewController/Classes/ConsentViewController.swift
```diff
@@ -271,29 +253,32 @@ import Reachability
         if(messageStatus == .loading || messageStatus == .loaded) { return }
 
         messageStatus = .loading
-        loadView()
+//        loadView()
         print ("url: \((messageUrl.absoluteString))")
         UserDefaults.standard.setValue(true, forKey: "IABConsent_CMPPresent")
-
-        setSubjectToGDPR { (optionalErrorObject) in
```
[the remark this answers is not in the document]
vilasma: currently not but we have to call GDPR somewhere and it will independent from webview load URL, once that is done will remove the GDPR call from here

Example/ConsentViewController/ViewController.swift
```diff
@@ -10,32 +10,41 @@ import UIKit
...
+        
+        let webview = WebView(frame: self.view.frame)
+        cvc.webview = webview
+        
+        cvc.onErrorOccurred = { error in
+            print(error)
+            webview.removeFromSuperview()
         }
-
+
```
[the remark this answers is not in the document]
vilasma: I will try on this while commenting the code to repo branch

ConsentViewController/Classes/WebView.swift
vilasma: will discuss on this when you have time over slack/call

ConsentViewController/Classes/WebView.swift
vilasma: Yeah we can, good suggestion.

ConsentViewController/Classes/WebView.swift
vilasma: we have our protocol methods implementation in ConsentViewController so we have to pass "self" of ConsentViewController not ViewControllers.
ViewController's "self" is required for "Webview" not for protocol.

PR #55, opened 2019-10-25
Carthage Support

vilasma:
Implemented changes for carthage into schem and created a tag 3.0.1 for carthage


vilasma:
@andresilveirah , I have added all the steps on Wiki page that to be carried out by our SDK user through Carthage. Can you have a look over that page so we can add all these steps to `README`.
https://github.com/SourcePointUSA/ios-cmp-app/wiki/Step-by-step-guide-for-Carthage

PR #56, opened 2019-11-05
Added webview as subview to viewcontroller Programmatically.

vilasma:
Now I am able to load the message on real devices.


andresilveirah:
No longer necessary with latest release

PR #60, opened 2019-11-07
Added clear consent data feature when we get onErrorOccurred callback event

vilasma:


ConsentViewController/Classes/ConsentViewController.swift
```diff
@@ -490,6 +490,19 @@ import Reachability
         didMessageScriptLoad = true
         consentDelegate.onMessageReady(controller: self)
     }
+    
+    /// It will clear all the stored userDefaults Data
+    public func clearAllConsnetData() {
```
andresilveirah: There's a typo on the name, `cosnet` instead of `consent`

vilasma: Chnaged typo from `cosent` to `consent`

PR #61, opened 2019-11-14
Added unit test cases for Consentviewcontroller SDK

vilasma:


SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentTest.swift
```diff
@@ -0,0 +1,30 @@
...
+    override func setUp() {
+        consent = Consent(id: "5b76ebcd5582409726cdd85", name: "Facebook")
+    }
+
+    override func tearDown() {
+    }
+    
+    /// this test method is used to test consent string method
+    func testConsentDescription() {
+        if !(consent?.description.isEmpty)! {
```
andresilveirah: I think here we could assert that `consent.description == "Consent(id: 5b76ebcd5582409726cdd85, name: Facebook"`.

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentViewControllerErrorTest.swift
```diff
@@ -0,0 +1,143 @@
...
+        
+    }
+    
+    /// this test method is used to test parsing error method
+    func testUnableToParseConsentStringError() {
+        let errorObject = UnableToParseConsentStringError(euConsent: "BOl9F8EOl9F8EAGABBENCi-AAAAp6ABAFIA")
+        if (errorObject.failureReason != nil) && !errorObject.description.isEmpty {
+            XCTAssert(true, "Error handling for the consent parsing failure is successfully handled")
+        } else {
+            XCTAssert(false, "Error handling for the consent parsing failure is not handled properly")
```
andresilveirah: We should test the actual content of the error messages. E.g. 
```
UnableToParseConsentStringError(euConsent: "BOl9F8EOl9F8EAGABBENCi-AAAAp6ABAFIA").description == "Could not parse the raw string BOl9F8EOl9F8EAGABBENCi-AAAAp6ABAFIA into a ConsentString"
```

the same is true for all other tests.

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentViewControllerTest.swift
```diff
@@ -0,0 +1,295 @@
...
+        consentViewController?.loadView()
+        if consentViewController?.webView != nil {
+            XCTAssert(true, "Webview initialized successfully")
+        } else {
+            XCTAssert(false, "Unable to initialize webview")
+        }
+    }
+    
+    /// this method is used to test whether message url is created or not successfully
+    func testGetMessageUrlWithAuthId() {
```
andresilveirah: I believe this belongs to `SourcePointClientTest` rather than `ConsentViewControllerTest`. 
Same for the function below `testGetMessageUrlWithoutAuthId`
Furthermore, we should assert upon its return, not only `!= nil`

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentViewControllerTest.swift
```diff
@@ -0,0 +1,295 @@
...
+                XCTAssert(false, "failed to get GDPR status from endpoint")
+            }
+            finished = true
+        }
+        while !finished {
+            RunLoop.current.run(mode: .default, before: Date.distantFuture)
+        }
+    }
+    
+    /// this method is used to test whether getIABVendorConsents method is able to get vendors or not from server
```
andresilveirah: * the comment is incorrect. All iAB-related consents are retrieved from the `UserDefaults` and not from the server.
* we should assert upon the real return value rather than `!= nil` only. This is valid for all functions below.

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/SourcePointClientTest.swift
```diff
@@ -0,0 +1,147 @@
...
+        }
+    }
+    
+    override func tearDown() {
+    }
+    
+    /// this test method is used to test getGdprStatus method working as expected or not
+    func testGetGdprStatus() {
+        var finished = false
+        sourcepointClient?.getGdprStatus { (gdprStatus, error) in
```
andresilveirah: * Can't we use `XCTestExpectation` for asynchronous testing?
* Is `print("Gdpr Status", gdprStatus ?? "Not found")` necessary?

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/SourcePointClientTest.swift
```diff
@@ -0,0 +1,147 @@
...
+        while !finished {
+            RunLoop.current.run(mode: .default, before: Date.distantFuture)
+        }
+    }
+    
+    /// this test method is used to test getMessageUrl method working as expected or not
+    func testGetMessageUrlWithAuthID() {
+        let targetingParamModel = ["TargetingParamModel": "false"]
+        let url = try! sourcepointClient?.getMessageUrl(forTargetingParams: targetingParamModel, debugLevel: "OFF", newPM: false, authId: "sourcepointtest")
+        if let urlString = url?.absoluteString {
```
andresilveirah: assert upon the real return value

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/SourcePointClientTest.swift
```diff
@@ -0,0 +1,147 @@
...
+            if urlString.isEmpty {
+                XCTAssert(false, "failed to create the url")
+            }else {
+                XCTAssert(true, "succeed to create the url")
+            }
+        }
+    }
+    
+    /// this test method is used to test getMessageUrl method working as expected or not
+    func testGetMessageUrlWithoutAuthID() {
```
andresilveirah: assert upon the real return value

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentViewControllerTest.swift
vilasma: In `ConsentViewController` class also we have methods where we are getting message URL.
We can use dynamic data for testing from `TestData.plist` file. So If I used assert upon return value and used account/property is different then test case will get fail.

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentTest.swift
vilasma: Done, If we change the format of description in future then we have to change test case also thats why I was just checking whether the description is empty or not. by this way also we can conclude that we are getting consent data or not.

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentViewControllerErrorTest.swift
vilasma: Done. If we change the format of error description in future then we have to change test cases also thats why I was just checking whether the description is empty or not.

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentViewControllerTest.swift
vilasma: Updated the comment.

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/SourcePointClientTest.swift
vilasma: 1. We can use either `XCTestExpectation` or ```while !finished {
            RunLoop.current.run(mode: .default, before: Date.distantFuture)
        }``` for  asynchronous testing both way we can achieve the result.

2. print statements is not required so I have removed that.

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentViewControllerTest.swift
vilasma: I have added assert on real value where it is possible

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/ConsentViewControllerTest.swift
vilasma: Asserts are used on real values

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/SourcePointClientTest.swift
vilasma: The URL which is generated with same account/property is different every time with the sequence. thats why here we have to check whether `urlString` is empty or not.

SourcePointMetaApp/SourcePointMetaAppTests/ConsentViewControllerTests/SourcePointClientTest.swift
vilasma: The URL which is generated with same account/property is different every time with the sequence. thats why here we have to check whether urlString is empty or not.

PR #64, opened 2019-11-21
Fix memory leak from WKWebView

andresilveirah:


vilasma:
@andresilveirah are we releasing webview(`releaseWebViewHandlers`) only if we encountered `onErrorOccured`. what if user completes interaction with webview?

andresilveirah:
@vilasma we're also calling on `.done()` https://github.com/SourcePointUSA/ios-cmp-app/pull/64/files#diff-76246d3a966ad7ce8b0bd0f75d46a94eR494

PR #66, opened 2019-11-26
Feature/rollbar analytics

vilasma:
Implemented the `Rollbar` analytics in SDK

### DIFF
--- a/ConsentViewController/Classes/ConsentViewController.swift
+++ b/ConsentViewController/Classes/ConsentViewController.swift
@@ -88,6 +88,9 @@ import Rollbar
     
     /// The variable is used to store the current SDK version which is used for analytics.
     static private let CONSENTVIEWCONTROLLER_SDK_VERSION = "3.0.0"
+    
+    /// The varibale used to set the flag to use Rollbar analytics
+    public var isDisableRollbar = false
 
     private var targetingParams: [String: Any] = [:]
     /// :nodoc:
@@ -506,6 +509,7 @@ import Rollbar
     
      /// This method is used to send the details to rollbar analytics about the error message with other deatils.
     internal func rollBarAnalytics(error: String) {
+        if !isDisableRollbar {
         let configuration = RollbarConfiguration()
         configuration.crashLevel = "critical"
         configuration.environment = "production"
@@ -518,6 +522,9 @@ import Rollbar
             "showPM": showPM,
             "messageTimeoutInSeconds": messageTimeoutInSeconds
         ])
+        }else {
+            logger.log("Rollbar is disabled by client application", [])
+        }
     }
         
     private func showMessage() {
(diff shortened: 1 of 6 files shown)


ConsentViewController/Classes/ConsentViewController.swift
```diff
@@ -62,10 +63,10 @@ import Reachability
     /// If the user is subject to GDPR, reading for this key in the `UserDefaults` will return "1" otherwise "0"
     static public let IAB_CONSENT_SUBJECT_TO_GDPR: String = "IABConsent_SubjectToGDPR"
 
-    /// They key used to store the IAB Consent string for the user in the `UserDefaults`
+    /// The key used to store the IAB Consent string for the user in the `UserDefaults`
     static public let IAB_CONSENT_CONSENT_STRING: String = "IABConsent_ConsentString"
 
-    /// They key used to read and write the parsed IAB Purposes consented by the user in the `UserDefaults`
```
andresilveirah: oops 😅

ConsentViewController/Classes/ConsentViewController.swift
```diff
@@ -84,6 +85,12 @@ import Reachability
     static public let IN_APP_MESSAGING_PAGE_DOMAIN = "in-app-messaging.pm.sourcepoint.mgr.consensu.org/v3/index.html"
 
     static private let MESSAGE_HANDLER_NAME = "JSReceiver"
+    
+    /// The variable is used to store the current SDK version which is used for analytics.
+    static private let CONSENTVIEWCONTROLLER_SDK_VERSION = "3.0.0"
+    
+    /// The varibale used to set the flag to use Rollbar analytics
+    public var isDisableRollbar = false
```
andresilveirah: I think it'd be better if we name it something a bit more generic (just in case we change service providers in the future). What about `isTelemetryEnabled`? Notice that with this new name, the value is no longer `false` but `true` by default.

ConsentViewController/Classes/ConsentViewController.swift
```diff
@@ -492,11 +501,32 @@ import Reachability
...
-
+    
+     /// This method is used to send the details to rollbar analytics about the error message with other deatils.
+    internal func rollBarAnalytics(error: String) {
+        if !isDisableRollbar {
+        let configuration = RollbarConfiguration()
+        configuration.crashLevel = "critical"
+        configuration.environment = "production"
+        Rollbar.initWithAccessToken("8c9341f5b0cd4701b43fd06237b0b660", configuration: configuration)
+        Rollbar.critical(withMessage: error, data: [
```
andresilveirah: By adding `Rollbar` directly in our code we're strongly coupling our code with it. 
We should implement a protocol and a wrapper around Rollbar. This is will make it easier to test and decouple our code from `Rollbar`.

vilasma:
The PR is currently targeting to `master` branch but as now we have `develop` branch so I will raise a PR to `develop` now.

PR #68, opened 2019-11-28
Feature/meta app migration to property

vilasma:
SourcePoint Meta app is moved into Example app of SDK as a target.


vilasma:
Meta app is moved into SDK Example app as a target and changed siteId to propertyId and siteName to Property into Meta app.
Going to merge this PR

PR #71, opened 2019-12-05
Feature/rollbar analytics

vilasma:
Added generic name to RollBar class and protocol method


andresilveirah:
@vilasma I'm closing this PR as we're no longer going to use Rollbar :/

PR #73, opened 2019-12-12
Implemented Connectivity Manager for network check

vilasma:
Implemented Connectivity Manager for network check and removed the Reachability 3rd party framework


andresilveirah:
Furthermore, since `ConnectivityManager` has no attributes, what if we change all functions to `static`? This way we don't need to have the overhead of an `init` and `shared` instance. What you think?

andresilveirah:
@vilasma I'm merging this PR as I will rebase #72 on top of it. Let's address the comments in another PR.

PR #75, opened 2020-01-09
Added GDPR prefix before the class/protocol names which are similar in both the sdks.

vilasma:


andresilveirah:
@vilasma I've applied your changes to a local branch because git was giving me headache. 
It should be same as #76

PR #82, opened 2020-02-12
Feature/carthage support

vilasma:
Added support for Carthage


andresilveirah:
@vilasma are the 2 first commits (28e50ff and fd56970) necessary for this PR? Could we cherry pick them away and keep only the Carthage related changes?

vilasma:
@andresilveirah Closing this PR and will raise another PR with Carthage changes.

PR #98, opened 2020-03-23
update wrapper api native-message url

andresilveirah:
Add missing query param `inApp=true` to `native-message` endpoint.


ConsentViewController/Classes/SourcePointClient.swift
```diff
@@ -85,7 +85,7 @@ A Http client for SourcePoint's endpoints
  */
 class SourcePointClient {
     static let WRAPPER_API = URL(string: "https://wrapper-api.sp-prod.net/tcfv2/v1/gdpr/")!
-    static let GET_MESSAGE_CONTENTS_URL = URL(string: "native-message", relativeTo: SourcePointClient.WRAPPER_API)!
```
leodoin: is this really the only code that had to be added?

PR #124, opened 2020-04-20
Added check for onPMReady JS callback

vilasma:


ConsentViewController/Assets/GDPRJSReceiver.js
```diff
@@ -63,7 +63,8 @@
           name: event.name,
           fromPM: event.fromPM,
```
leodoin, line 64: here you could write something like this:
```javascript
fromPM: isFromPm(event)

function isFromPm(event){
  //ninja code
  return event.fromPM != null? event.fromPM : (event.settings.vendorList == null);
}
```

ConsentViewController/Assets/GDPRJSReceiver.js
```diff
@@ -63,7 +63,8 @@
...
           actionType: event.actionType,
-          payload: event.payload || event.actions || {}
+          payload: event.payload || event.actions || {},
+          vendorList: event.settings.vendorList
```
leodoin, line 67: basically it looks like you should be getting a nullPointerException

Example/ConsentViewController/ViewController.swift
```diff
@@ -37,6 +37,14 @@ extension ViewController: GDPRConsentDelegate {
     func consentUIDidDisappear() {
         dismiss(animated: true, completion: nil)
     }
+
```
leodoin, line 40: why is this commented out?

PR #125, opened 2020-04-21
Added fromPM check for onPMReady SDK call 

vilasma:


ConsentViewController/Assets/GDPRJSReceiver.js
```diff
@@ -61,7 +61,7 @@
       try {
         handleMessageEvent(SDK)({
           name: event.name,
-          fromPM: event.fromPM,
+          fromPM: event.fromPM != null ? event.fromPM : (event.settings.vendorList != null),
```
leodoin, line 64: I think it is better to encapsulate this in a function like
```javascript
function isFromPm(event){
  return event.fromPM != null ? event.fromPM : (event.settings.vendorList != null)
}
```
or maybe i'm being too much of a clean code evangelist 😅

vilasma: Done @leodoin

PR #132, opened 2020-04-27
Disable zoom in scrollview

vilasma:


ConsentViewController/Classes/MessageWebViewController.swift
```diff
@@ -223,4 +224,12 @@ extension MessageWebViewController: UIScrollViewDelegate {
     func viewForZooming(in scrollView: UIScrollView) -> UIView? {
         return nil
     }
+
+    func disableZoomInScrollView() -> WKUserScript {
+        let source: String = "var meta = document.createElement('meta');" +
```
andresilveirah, line 229: I've tried loading a dummy html page with the following tag:
```html
<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum- scale=1.0, user-scalable=no">
```

Just like the one you're creating on the fly in this method. 
Unfortunately that didn't solve the issue. Are you confident the issue is solved by adding this to webview? If that's the case, we can also ask the Web team to add that meta tag to the message/PM web page, so we don't have to add extra code to our SDK.

vilasma: @andresilveirah I have tested many times with pinch zoom and rotation and its working as expected. can you give a try with your device as on simulator its works smooth.

andresilveirah: @vilasma , Matt has deployed a change adding the above to the `notice` app. If that works, we won't need to add this code to the SDK.

andresilveirah:
@vilasma since this was fixed in the "web" side, I believe we can close this issue.

PR #133, opened 2020-04-27
Fix consent delegate methods v4

andresilveirah:
* renamed `pmWillShow/DidDisappear` to `gdprPMWillShow/DidDisappear`
* fixed `gdprPMWillShow/DidDisappear` not firing

### DIFF
--- a/ConsentViewController/Assets/GDPRJSReceiver.js
+++ b/ConsentViewController/Assets/GDPRJSReceiver.js
@@ -42,31 +42,39 @@
   }
 
   var handleMessageEvent = function(SDK) {
-    return function(name, payload) {
-      switch(name) {
+    return function(eventData) {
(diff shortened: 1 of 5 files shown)


ConsentViewController/Assets/GDPRJSReceiver.js
```diff
@@ -42,31 +42,39 @@
...
-          payload.action = name;
-          SDK.onMessageEvent(payload);
+          eventData.payload.action = eventData.name;
+          SDK.onMessageEvent(eventData.payload);
       }
     }
   };
 
+  function isFromPM(event) {
+    return event.settings == null
```
vilasma, line 63: @andresilveirah in GDPR we have to check **vendorList**attribute and in CCPA we are checking **settings** attribute

PR #136, opened 2020-05-04
Removed strong reference of OnError

vilasma:
Removed strong reference of OnError, onwards passing error in completion handler also removed shared instance of Connectivity Manager. updated unit test cases for the same.

### DIFF
--- a/ConsentViewController/Classes/MessageWebViewController.swift
+++ b/ConsentViewController/Classes/MessageWebViewController.swift
@@ -140,7 +140,8 @@ class MessageWebViewController: GDPRMessageViewController, WKUIDelegate, WKNavig
     }
 
     func load(url: URL) {
-        if ConnectivityManager.shared.isConnectedToNetwork() {
+        let connectvityManager = ConnectivityManager()
+        if connectvityManager.isConnectedToNetwork() {
             webview?.load(URLRequest(url: url))
         } else {
             onError(error: NoInternetConnection())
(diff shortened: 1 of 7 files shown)


ConsentViewController/Classes/SourcePointClient.swift
```diff
@@ -130,8 +127,6 @@ class SourcePointClient {
     private let campaignEnv: GDPRCampaignEnv
     private let targetingParams: TargetingParams?
 
-    public var onError: OnError? { didSet { client.defaultOnError = onError } }
```
andresilveirah, line 133: Have you tried using an `unowned` reference instead of removing it?

vilasma: Yes @andresilveirah I have tried with both  **weak and unowned**  but both gives error.

ConsentViewController/Classes/MessageWebViewController.swift
```diff
@@ -140,7 +140,8 @@ class MessageWebViewController: GDPRMessageViewController, WKUIDelegate, WKNavig
     }
 
     func load(url: URL) {
-        if ConnectivityManager.shared.isConnectedToNetwork() {
+        let connectvityManager = ConnectivityManager()
```
andresilveirah, line 143: What about having the `conectivityManager` as an attribute of `GDPRConsentViewController` class? This way we don't need to instantiate it every time.

vilasma: We are using connectivity manager only once except SourcePointClient class, so instead of creating a global variable I have created a local variable.
If we are using that variable multiple times then we can create global one but in this case I feel we can keep this local one so it will exist only in that method and we don't have to take care outside of method.

PR #139, opened 2020-05-13
SP-3524 implement customConsent API

andresilveirah:
It's now possible to add vendors, categories and legitimateInterestCategories consents to a given consentUUID. In other words, programatically consent an user to a list of vendors/cat/legIntCat.

* Updated SourcePointClient to use the new `custom-consent` endpoint
* Updated GDPRConsentViewController to expose `customConsentTo` method
* Updated the Example app to show off the new feature

### DIFF
--- a/Example/ConsentViewController/ViewController.swift
+++ b/Example/ConsentViewController/ViewController.swift
@@ -10,6 +10,9 @@ import UIKit
 import ConsentViewController
 
 class ViewController: UIViewController {
+    static let vendorXId = "5e4a5fbf26de4a77922b38a6"
+    var vendorXAccepted = false
+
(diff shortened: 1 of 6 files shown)


Example/ConsentViewController/ViewController.swift
```diff
@@ -28,18 +31,38 @@ class ViewController: UIViewController {
...
+
+    @IBOutlet weak var vendorXStatusLabel: UILabel!
+    
     override func viewDidLoad() {
         super.viewDidLoad()
         consentViewController.loadMessage()
     }
 }
 
+// MARK: CustomConsent
```
leodoin, line 57: should we keep this changes on the example app?

andresilveirah: I'd like to @leodoin although the UI doesn't look good...

PR #168, opened 2020-06-08
Initial automation test suite for TCFv2

deshpandevrushali:


andresilveirah:
@vilasma @deshpandevrushali 
1. can the folder `.metadata` be added to the gitignore?
2. there are a couple of `.jar` and `.exe` files, can we ignore the binaries as well?
3. I notice we have a file called `FusionLicenceRegistration_v3.3.jar`. Are you sure we're not including any sensitive data in this PR? Like the key to Fusion?

PR #184, opened 2020-06-11
Ensured onConsentReady is always called at PM cancel event

vilasma:


ConsentViewController/Classes/GDPRConsentViewController.swift
```diff
@@ -374,15 +374,6 @@ extension GDPRConsentViewController: GDPRConsentDelegate {
                     self?.onError(error: error)
                 }
             }
-        } else if (action.type == .Dismiss) {
```
andresilveirah, line 377: @vilasma why was this code deleted? If the message is dismissed, will the `onConsentReady` still be called?

vilasma: @andresilveirah Yes, I moved the dismiss action call to MessageWebViewController.

PR #190, opened 2020-06-16
SP-3631 always call on consent ready

andresilveirah:
Ensure `onConsentReady` is always called after every action with exception of:
1. `ShowPrivacyOptions`
2. `PMCancel` when in the 2nd layer message.


andresilveirah:
Although the automation scripts are failing on Jenkins, I've ran them locally and they passed (apart from 2 known issues). I'm going ahead and merging this branch.

PR #191, opened 2020-06-16
SP-3889 Added cachePolicy with ReloadIgnoringCacheData to URLSession.

vilasma:


andresilveirah:
@vilasma try covering your change in the unit tests as well. It should be one extra `it` asserting the configuration was set to `.reloadIgnoringLocalCacheData`.

PR #200, opened 2020-06-26
SP-4022 start stop appium from code

deshpandevrushali:


AutomationFramework/src/main/java/tests/GDPR_TCFV2_MetaAppTests.java
```diff
@@ -215,7 +219,7 @@ public void CheckConsentOnAcceptAllFromConsentView() throws InterruptedException
 					"Expected consent message not displayed");
 
 			mobilePageWrapper.consentViewPage.scrollAndClick("ACCEPT ALL");
-
+			System.out.println(mobilePageWrapper.siteDebugInfoPage.GDPRConsentUUID.getText());
```
andresilveirah, line 222: @deshpandevrushali should we use the `logMessage` method for logging?

AutomationFramework/src/main/java/tests/GDPR_TCFV2_MetaAppTests.java
```diff
@@ -330,7 +335,7 @@ public void CheckConsentOnRejectAllFromPM() throws InterruptedException, NoSuchE
...
```
andresilveirah, line 338: Why do we need to log the `consentUUID` ? Debugging?

AutomationFramework/src/main/java/tests/GDPR_TCFV2_MetaAppTests.java
deshpandevrushali: @andresilveirah I am started working on SP-4077 reporting/logging for framework so that we can remove this explicitly Sysout statements from tests. https://sourcepoint.atlassian.net/browse/SP-4077

deshpandevrushali: @andresilveirah by using this sysout for now we are waiting explicitly for respective elements to load and currently we have only one test failure which is known and all others are passing.

andresilveirah: @deshpandevrushali still not sure why we have the `sysout` statements there? I've checked [SP-4077](https://sourcepoint.atlassian.net/browse/SP-4077) but the description is empty.

> I have started working on SP-4077 reporting/logging for framework 
What reporting/logging are you talking about?

deshpandevrushali: @andresilveirah I have added description for SP-4077. with this ticket we are implementing logging system for automation test execution to avoid sysout also will get proper logging in place so that we can have execution details on test failure.
Reporting consists of implementing Allure reporting framework to get HTML test execution report.

andresilveirah: Got it. So the logging we want to implement will give us better insights when something fails.

PR #201, opened 2020-06-29
[SP-4024] Get pm id from action

vilasma:
- Get PM url from JSReceiver
- Extract PM Id from PM url


Example/ConsentViewController_ExampleTests/MessageWebViewControllerSpec.swift
```diff
@@ -363,6 +338,15 @@ class MessageWebViewControllerSpec: QuickSpec, GDPRConsentDelegate, WKNavigation
...
 
+        describe("getPMIdFromMessage") {
+            it("returns pmId from PM url") {
+                let payload = "{\"pm_url\":\"https://notice.sp-prod.net/privacy-manager/index.html?message_id=122058\"}".data(using: .utf8)!
+                let action = GDPRAction(type: .AcceptAll, id: "something", payload: payload)
+                messageWebViewController.getPMIdFromMessage(action: action)
+                expect(messageWebViewController.pmId).to(equal("122058"))
+            }
+        }
+
```
andresilveirah, line 349: @vilasma I suggest we avoid testing "protected" methods and try to test the behaviour via the public ones. Example, instead of testing `getPMIdFromMessage`, we test the `userContentController` receiving a _Show Privacy Manager_ action with the url in the payload. 
That might require a bit more boilerplate to write the test, but will enable us to refactor the code in the future without having to re-write the specs.

vilasma: @andresilveirah I have added test case to check PM Id extraction from pm_url using `userContentController`.

PR #203, opened 2020-06-29
SP-4077 implement reporting logging

deshpandevrushali:


deshpandevrushali:
> @deshpandevrushali should the automation job pass in Jenkins or is this expected to fail?

@andresilveirah automation job expected to fail with 1 test failure. If there are more than that I will look into it. With following job status yesterday we got 1 failure.
http://sdk-jenkins.sourcepoint.com/job/iOS_Automation_GDPR_PR/215/console

PR #205, opened 2020-07-02
SP-4076 add ui tests

vilasma:


Example/SPGDPRExampleAppUITests/SPGDPRExampleAppUITests.swift
```diff
@@ -20,59 +20,143 @@ class SPGDPRExampleAppUITests: QuickSpec {
             self.app = ExampleApp(XCUIApplication())
             self.app.launch()
             self.app.clearConsentButton.tap()
+            Nimble.AsyncDefaults.Timeout = 10
+            Nimble.AsyncDefaults.PollInterval = 1
         }
 
-        afterSuite {
+        afterEach {
```
andresilveirah, line 27: I believe we shouldn't run the following two lines in the `afterEach` block:
```swift
Nimble.AsyncDefaults.Timeout = 1
Nimble.AsyncDefaults.PollInterval = 0.01
```
That might be one of the reasons the tests are failing. 

Try moving them to an `afterSuit` block.

Example/SPGDPRExampleAppUITests/SPGDPRExampleAppUITests.swift
```diff
@@ -20,59 +20,143 @@ class SPGDPRExampleAppUITests: QuickSpec {
...
+            Nimble.AsyncDefaults.PollInterval = 0.01
         }
 
         describe("SPGDPRExampleAppUITests") {
-            context("when launching the app for the first time") {
-                it("1. should see the consent message") {
-                    expect(self.app.consentMessage).to(showUp())
+            it("1. Accept all through message") {
+                expect(self.app.consentMessage).to(showUp())
+                if self.app.consentMessage.exists {
```
andresilveirah, line 39: @vilasma is it necessary to call `swipeUp()`? The tests seem to pass without it. 
If we have to use it, do you see a way to "hide" them inside the `ExampleApp` class?

PR #212, opened 2020-07-15
[SP-4181] add ui tests for PM

vilasma:


Example/SPGDPRExampleAppUITests/SPGDPRExampleAppUITests.swift
```diff
@@ -124,6 +124,15 @@ class SPGDPRExampleAppUITests: QuickSpec {
             self.app.saveAndExitButton.tap()
             expect(self.app.privacyManager).to(disappear())
         }
+
+        it("Terms And Conditions link opens in other webview") {
```
andresilveirah, line 128: @vilasma it should open on Safari instead. Isn't this the case?

vilasma: @andresilveirah link open on other Safari webview only. should I add `Safari` word also?

andresilveirah: @vilasma I believe it'd be more explicit if we exchanged _webview_ for _Safari_ or _Browser_.

PR #218, opened 2020-08-03
Fix issue when GDPRMessage didn't have all its fields present when JSON encoded

ivanlisovyi:
## Context 
This PR fixes an issue when `GDPRMessage` -> `JSON` encoding produced a JSON without `choiceType` and `choiceId` fields for the `MessageAttributeAction` objects. 

## Affected Functionality 
- `GDPRMessage` JSON Encoding

### DIFF
--- a/ConsentViewController/Classes/GDPRMessage.swift
+++ b/ConsentViewController/Classes/GDPRMessage.swift
@@ -56,6 +56,14 @@ public typealias CustomFields = [String: String]
         )
     }
 
+    public override func encode(to encoder: Encoder) throws {
+        try super.encode(to: encoder)
+
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(choiceId, forKey: .choiceId)
+        try container.encode(choiceType, forKey: .choiceType)
+    }
+
     enum CodingKeys: String, CodingKey {
         case text, style, customFields, choiceId, choiceType
     }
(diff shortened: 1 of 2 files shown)


andresilveirah:
Hi @ivanlisovyi thank you for your PR (I mean, it even covers the fix with specs)! I'd like to understand when do you face the case in which the `MessageAttributeAction` does not contain `choiceType` and `choiceId`. I want to make sure we're to pass this along to our backend colleagues as it might be a problem with the API.

ivanlisovyi:
@andresilveirah 

> Hi @ivanlisovyi thank you for your PR (I mean, it even covers the fix with specs)! I'd like to understand when do you face the case in which the `MessageAttributeAction` does not contain `choiceType` and `choiceId`. I want to make sure we're to pass this along to our backend colleagues as it might be a problem with the API.

It's not a problem with the backend API, when we get a `GDPRMessage` from the SDK we try to convert it back to a json using the `JSONDecoder` and the resulting json doesn't have mentioned fields. 

This happens because `MessageAction` inherits implementations of `init(decoder:)` and `encode(with:)` from the `MessageAttribute` and overrides only `init(decoder:)` which essentially means that `MessageAction` will use parent implementation of the `encode(with:)` function when we try to encode it with `JSONEcoder`.

andresilveirah:
@ivanlisovyi that makes absolute sense. Thank you again for the PR.

PR #220, opened 2020-08-06
refactor example app ui tests

andresilveirah:
The ui tests for the Example app is failing (sometimes) https://github.com/SourcePointUSA/ios-cmp-app/actions/runs/197651154/workflow

I suspect it's due to the asynchronous nature of the `customConsentTo` feature.


vilasma:
Yes @andresilveirah, sometimes `customConsentTo` test fails in that case by re-running the job we can pass the test job.
I think changes are not required

andresilveirah:
I'm not sure @vilasma . It's failing quite consistently on GitHub but it passes locally. I'll keep looking into it.

PR #224, opened 2020-08-10
[SP-4457] Fix authenticated consent not working issue

vilasma:


ConsentViewController/Classes/GDPRConsentViewController.swift
```diff
@@ -263,7 +265,7 @@ typealias Meta = String
 
     func didAuthIdChange(newAuthId: String?) -> Bool {
         return newAuthId != nil &&
-            localStorage.authId != nil &&
+            localStorage.authId?.count ?? 0 > 0 &&
```
andresilveirah, line 268: @vilasma I think we could use 
```swift
localStorage.authId?.isEmpty ?? true
```

ConsentViewController/Classes/GDPRConsentViewController.swift
```diff
@@ -221,11 +221,13 @@ typealias Meta = String
                 clearAllData()
                 localStorage.authId = authId
             }
+            localStorage.authId = authId
             sourcePoint.getMessage(native: native, consentUUID: gdprUUID, euconsent: euconsent, authId: authId, meta: localStorage.meta) { [weak self] messageResponse, error in
                 if let messageResponse = messageResponse {
                     self?.localStorage.consentUUID = messageResponse.uuid
                     self?.localStorage.meta = messageResponse.meta
                     self?.localStorage.userConsents = messageResponse.userConsent
+                    self?.localStorage.authId = authId
```
andresilveirah, line 230: I believe we don't need to add the authId to the localStorage again here.

vilasma: Forgot to remove, will remove with new commint

ConsentViewController/Classes/GDPRConsentViewController.swift
vilasma: @andresilveirah this will always return true instead of that will try with `localStorage.authId?.isEmpty ?? false`

PR #228, opened 2020-08-19
[SP-4468] Fix - Webview is no longer fullscreen after dismissing the keyboard

vilasma:


ConsentViewController/Classes/MessageWebViewController.swift
```diff
@@ -70,9 +70,17 @@ class MessageWebViewController: GDPRMessageViewController, WKUIDelegate, WKNavig
     }
 
     override func loadView() {
+        NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillHide), name: UIResponder.keyboardWillHideNotification, object: nil)
```
andresilveirah, line 73: @vilasma let's add a comment here explaining we need to add this piece of logic because of an bug on iOS 12/13 regarding the Keyboard shifiting the view up after hiding.

andresilveirah (approved):
Very elegant solution @vilasma 💯

Issue #229, opened 2020-08-20
[SPRING] Non-Network related errors while loading Privacy Manager URL (and any URL in general) in the webview are ignored 


ivanlisovyi:
**Describe the bug**
When loading of the privacy manager fails because of some problems with SP backend (e.g. 500), the SDK doesn't call `onError` delegate method. 

**About your campaign (please complete the following information):**
Account Id 75
Property Id 4598
Property Name "WELT-NEWS-IOS"
PMId "134094"
Campaign Env

**About your environment**
SDK version 5.2.6
Device: iPhone 8
OS: 13.6

**To Reproduce**
1. Map the Privacy Manager URL (https://notice.sp-prod.net/privacy-manager/*) response and return 500 status code in the Proxy app of your choice (Charles, Proxyman, etc) 
2. Present `GDRPConsentViewController` and try to `loadPrivacyManager` 
3. Nothing happens until timeout occurs

**Expected behavior**
When loading of the Privacy Manager URL fails because of non-network related reason (e.g. 500) the SDK shall report the error through the `onError` delegate method 

**Additional Info** 
I've created a PR with a quick fix for this issue https://github.com/SourcePointUSA/ios-cmp-app/pull/230 but feel free to ignore or close it as long as the issue is resolved.

andresilveirah:
@ivanlisovyi the fix is now live in the latest release `5.2.8` 💯

PR #230, opened 2020-08-20
#229 Make sure that SDK calls onError delegate when webView fails to load url

ivanlisovyi:
This PR fixes the following issue https://github.com/SourcePointUSA/ios-cmp-app/issues/229


andresilveirah:
@ivanlisovyi you're too fast too furious! We started working in this EXACT issue this morning!

andresilveirah:
@ivanlisovyi the changes are working like a charm, thank you for your contribution.

PR #232, opened 2020-08-26
[SP-4723] Dark mode support to MetaApp

vilasma:


Example/SourcePointMetaApp/Controllers/PropertyListViewController/PropertyListViewController.swift
```diff
@@ -173,7 +173,7 @@ extension PropertyListViewController: UITableViewDelegate {
             if let editpropertyViewController = UIStoryboard(name: "Main", bundle: nil).instantiateViewController(withIdentifier: "AddPropertyViewController") as? AddPropertyViewController {
                 editpropertyViewController.propertyManagedObjectID = self.propertyListViewModel.propertyManagedObjectID(atIndex: indexPath.row)
 
-                self.navigationController!.pushViewController(editpropertyViewController, animated: true)
```
andresilveirah, line 176: @vilasma why did you decide to set the `animated` param to false? Was it causing issues?

vilasma: Yes @andresilveirah, when we move from one controller to other it moves with animation and for fraction of second we can see the previous controller components which are going to background. 
I have attached videos in ticket where we can see the animation effect in first video and in second without animation.

andresilveirah: I see the issue. But I think we should fix the problem rather than working around it by removing screen transitions.

vilasma: @andresilveirah,  When we push controller to navigation the default animation is set to true.
If we don’t want animation while screen transition then we simply set the animation to false.

`In iOS 7 Apple added a new default navigation behavior. You can swipe from the left edge of the screen to go back on the navigation stack.`  we can disable this feature from Meta App but it will work only in backward direction not in forward direction.

When we need some specific transition then we use `UIViewControllerAnimatedTransitioning` class for animation but as such we don’t have any need of animation in MetaApp.

PR #233, opened 2020-08-28
[SP-4113] implement publisherData on GDPRAction class

andresilveirah:
The publisher can now add their own arbitrary data to the action payload. 
```swift
// GDPRConsentDelegate
    func onAction(_ action: GDPRAction) {
        action.publisherData = ["my data": try? SPGDPRArbitraryJson("foo bar")]
    }
```
This will be encoded into:
```json
{
  ...,
  "pubData": {
     "my data": "foo bar"
  }
}
```
and sent to our endpoints / logs. Where it can then be retrieved via APIs.

### DIFF
--- a/ConsentViewController/Classes/GDPRAction.swift
+++ b/ConsentViewController/Classes/GDPRAction.swift
@@ -34,9 +34,10 @@ import Foundation
     public let type: GDPRActionType
     public let id: String?
     public let payload: Data
+    public var publisherData: [String: SPGDPRArbitraryJson?]?
 
     public override var description: String {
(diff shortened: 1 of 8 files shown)


vilasma (approved):

vilasma:
@andresilveirah we can add  
`  func onAction(_ action: GDPRAction) {
        action.publisherData = ["my data": try? SPGDPRArbitraryJson("foo bar")]
    }` 
in example app for our users as a use case, what you think?

andresilveirah:
@vilasma I haven't added that to our example app, just because this feature is not very used... So I created a wiki instead https://github.com/SourcePointUSA/ios-cmp-app/wiki/Sending-arbitrary-data-when-the-user-takes-an-action.

PR #243, opened 2020-09-30
Rename PMConsents to GDPRPMConsents

vilasma:


andresilveirah (commented):
@vilasma I believe we should bump the minor version with this code since it affects the SDK's public interface (`GDPRConsentDelegate`). 

Can you prepare the release in this same PR?

vilasma:
> @vilasma I believe we should bump the minor version with this code since it affects the SDK's public interface (`GDPRConsentDelegate`).
> 
> Can you prepare the release in this same PR?

vilasma:
@andresilveirah Released SDK with version 4.2.0 on CocoaPods and Carthage. 
Also this SDK is compiled with Xcode 12.0.1 so we are good with compatibility with Xcode 12 and iOS 14.

PR #254, opened 2020-10-20
[SP-5296] Replace wrapper url with CDN url

vilasma:


Example/ConsentViewController_ExampleTests/MessageWebViewControllerSpec.swift
```diff
@@ -120,7 +120,7 @@ class MessageWebViewControllerSpec: QuickSpec, GDPRConsentDelegate, WKNavigation
                         ])
 
                         messageWebViewController.userContentController(userContentController, didReceive: message)
-                        let expectedPMURL = "https://notice.sp-prod.net/privacy-manager/index.html?site_id=1&consentUUID=uuid&message_id=122058"
```
andresilveirah, line 123: @vilasma I think we could change this spec. The `WebView` doesn't "care" if the URL is `notice.sp-prod` or the CDN one. All we should test is that the query param `message_id` contains the right value (PM ID)

Example/ConsentViewController_ExampleTests/MessageWebViewControllerSpec.swift
```diff
@@ -133,7 +133,7 @@ class MessageWebViewControllerSpec: QuickSpec, GDPRConsentDelegate, WKNavigation
                             "body": ["type": 12, "id": "id", "payload": ["pm_url": "pm_url"]]
                         ])
                         messageWebViewController.userContentController(userContentController, didReceive: message)
-                        let expectedPMURL = "https://notice.sp-prod.net/privacy-manager/index.html?site_id=1&consentUUID=uuid&message_id=1234"
```
andresilveirah, line 136: Same here. We should check only for the query param, not the entire URL. Furthermore, I think it'd be more descriptive if we changed:
```swift
it("calls the onAction on the consent delegate with ShowPrivacyManager with wrong PM URL")
```
to
```swift
context("and the pm_url attribute is invalid") {
  it("uses its own pmId as message_id" {
     // test here
  }
}
```
This way, the `context` gives the pre condition, and the `it` explains what should happen.

Example/ConsentViewController_ExampleTests/MessageWebViewControllerSpec.swift
vilasma: @andresilveirah updated the spec.

Example/ConsentViewController_ExampleTests/MessageWebViewControllerSpec.swift
vilasma: @andresilveirah updated the spec.

Example/ConsentViewController_ExampleTests/MessageWebViewControllerSpec.swift
```diff
@@ -110,31 +110,34 @@ class MessageWebViewControllerSpec: QuickSpec, GDPRConsentDelegate, WKNavigation
...
+                            let webviewMock = WebViewMock()
+                            messageWebViewController.connectivityManager = ConnectivityMock(connected: true)
+                            messageWebViewController.webview = webviewMock
+                            let message = MessageMock([
+                                "name": "onAction",
+                                "body": ["type": 12, "id": "id", "payload": ["pm_url": "pm_url"]]
+                            ])
+                            messageWebViewController.userContentController(userContentController, didReceive: message)
+                            let expectedPMURL = "https://cdn.privacy-mgmt.com/privacy-manager/index.html?site_id=1&consentUUID=uuid&message_id=1234"
+                            expect(webviewMock.loadCalledWith.url?.absoluteString).to(equal(expectedPMURL))
```
andresilveirah, line 139: Here, instead of checking the full URL, we can simply check the `message_id` has the value `1234`. The result is the same, but it illustrates a little bit more explicitly what the functionality is doing (changing the `message_id`).

vilasma: @andresilveirah updated the expectation.

andresilveirah:
Cool @vilasma . I've asked Zeke and we need to hold back a bit to merge these changes. I'll let you know once we're good to go.

PR #260, opened 2020-11-04
[SP-5244] Add support for consent language

vilasma:
1. In Web message case we are sending **consentLanguage** to sendConsent API call, consentLanguage is captured from action event.
1. In Native message case we are sending **Locale** value as consentLanguage value to sendConsent API call.

### DIFF
--- a/ConsentViewController/Classes/GDPRAction.swift
+++ b/ConsentViewController/Classes/GDPRAction.swift
@@ -33,7 +33,7 @@ import Foundation
 @objcMembers public class GDPRAction: NSObject {
     public let type: GDPRActionType
     public let id: String?
-    public var consentLanguage: String?
+    public let consentLanguage: String?
     public let payload: Data
     public var publisherData: [String: SPGDPRArbitraryJson?] = [:]
 
--- a/ConsentViewController/Classes/GDPRAction.swift
+++ b/ConsentViewController/Classes/GDPRAction.swift
@@ -33,11 +33,15 @@ import Foundation
 @objcMembers public class GDPRAction: NSObject {
     public let type: GDPRActionType
     public let id: String?
+    public let consentLanguage: String?
     public let payload: Data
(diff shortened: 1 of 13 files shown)


ConsentViewController/Classes/SourcePointClient/SourcePointClient.swift
```diff
@@ -182,7 +182,9 @@ class SourcePointClient: SourcePointProtocol {
                 requestUUID: requestUUID,
                 pmSaveAndExitVariables: pmPayload,
                 meta: meta,
-                publisherData: action.publisherData
+                publisherData: action.publisherData,
+                consentLanguage: (action.consentLanguage != nil) ?
+                    action.consentLanguage : Locale.preferredLanguages[0].uppercased()
```
andresilveirah, line 187: @vilasma I think we shouldn't put the responsibility to fill in the default language `Locale.preferredLanguages[0].uppercased()` in the `SPClient`. 
This should be set as soon as the `Action` is created.

vilasma: @andresilveirah added Locale value in once action is triggered from native message

ConsentViewController/Classes/GDPRAction.swift
```diff
@@ -33,7 +33,7 @@ import Foundation
 @objcMembers public class GDPRAction: NSObject {
     public let type: GDPRActionType
     public let id: String?
-    public let consentLanguage: String?
+    public var consentLanguage: String?
```
andresilveirah, line 36: @vilasma does `consentLanguage` need to be a `var`? Are we reassigning it at some point?

vilasma: Yes @andresilveirah, In native message case we are assigning the `Locale` value to `consentLanguage`.

vilasma: I have updated consentLanguage property type to let

PR #267, opened 2020-12-04
[SP-5447] Update Meta App with message language

vilasma:


andresilveirah:
@vilasma here's my feedback regarding the message language picker view: 
1. noticed the user is allowed to write any content in the Message Language Input. The user shouldn't be allowed to type there.
2. if tapping on anywhere other than the PickerView, the language PickerView should be dismissed. 
3. There should be a label next to the message language input saying: "Message Language:", like we have with all other inputs.
4. the Browser Default is selected by default but it's not visible to the user (unless the user taps on the input).

Screenshot 2020-12-09 at 18 02 25

vilasma:
> @vilasma here's my feedback regarding the message language picker view:
> 
> 1. noticed the user is allowed to write any content in the Message Language Input. The user shouldn't be allowed to type there.
> 2. if tapping on anywhere other than the PickerView, the language PickerView should be dismissed.
> 3. There should be a label next to the message language input saying: "Message Language:", like we have with all other inputs.
> 4. the Browser Default is selected by default but it's not visible to the user (unless the user taps on the input).
> 
> Screenshot 2020-12-09 at 18 02 25

Thank you @andresilveirah for the feedback on picker view.
I have started working on feedback changes

PR #268, opened 2020-12-09
[SP-5623] Fix Consent persistence issue in PM as first layer message case

vilasma:


andresilveirah:
@vilasma let's wait a bit to merge this PR. While refactoring the error classes on the SDK I've also worked on the way we're dealing with query params. Let's first merge that and merge this on top afterwards.

andresilveirah:
Merge #269 before this one.

andresilveirah (approved):

PR #272, opened 2020-12-23
SP-5415 send error metrics

andresilveirah:
* refactor `GDPRConsentDelegate.onError` to receive non-optional Error parameter
* make sure to call the `GDPRConsentViewController.onError` on every error case
* send error metrics to wrapper api

### DIFF
--- a/ConsentViewController/Classes/SPDeviceManager.swift
+++ b/ConsentViewController/Classes/SPDeviceManager.swift
@@ -13,7 +13,7 @@ protocol SPDeviceManager {
 }
 
 struct SPDevice: SPDeviceManager {
-    /// Returns he Version of the OS. E.g 1.2
+    /// Returns the Version of the OS. E.g 1.2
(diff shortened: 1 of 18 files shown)


ConsentViewController/Classes/SPDeviceManager.swift
```diff
@@ -0,0 +1,35 @@
...
+
+import Foundation
+
+protocol SPDeviceManager {
+    func osVersion () -> String
+    func deviceFamily () -> String
+}
+
+struct SPDevice: SPDeviceManager {
+    /// Returns he Version of the OS. E.g 1.2
```
vilasma, line 16: Its typo instead of `he` we should change to `the

vilasma (approved):

ConsentViewController/Classes/SPDeviceManager.swift
andresilveirah: thanks @vilasma 👍

PR #273, opened 2020-12-30
[SP-5909] Implemented PMTab feature

vilasma:
- With this feature user can request pm tab from SDK.
- To have the requested pmTab we need to set the `.privacyManagerTab` attribute of the GDPRConsentViewController before calling `.loadMessage() / .loadPrivacyManager()`.
`consentViewController.privacyManagerTab = .Vendors`                                    
  `consentViewController.loadMessage()`

- Added pmTab feature in Meta App

### DIFF
--- a/ConsentViewController/Classes/PrivacyManagerTab.swift
+++ b/ConsentViewController/Classes/PrivacyManagerTab.swift
@@ -10,8 +10,8 @@ import Foundation
 /// Supported privacy manager tabs in PM
 public enum PrivacyManagerTab: String, CaseIterable {
     /// having it empty string instructs the rendering app to use the defult tab of PM
-    case DefaultTab = ""
-    case PurposesTab = "purposes"
-    case VendorsTab = "vendors"
-    case FeaturesTab = "features"
+    case Default = ""
+    case Purposes = "purposes"
+    case Vendors = "vendors"
+    case Features = "features"
 }
--- a/ConsentViewController/Classes/PrivacyManagerTab.swift
+++ b/ConsentViewController/Classes/PrivacyManagerTab.swift
@@ -1,0 +1,17 @@
+//
+//  PrivacyManagerTab.swift
+//  ConsentViewController
+//
+//  Created by Vilas on 28/12/20.
+//
+
+import Foundation
+
+/// Supported privacy manager tabs in PM
+public enum PrivacyManagerTab: String, CaseIterable {
+    /// having it empty string instructs the rendering app to use the defult tab of PM
+    case DefaultTab = ""
+    case PurposesTab = "purposes"
+    case VendorsTab = "vendors"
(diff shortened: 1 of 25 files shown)


ConsentViewController/Classes/PrivacyManagerTab.swift
```diff
@@ -0,0 +1,17 @@
...
+//  ConsentViewController
+//
+//  Created by Vilas on 28/12/20.
+//
+
+import Foundation
+
+/// Supported privacy manager tabs in PM
+public enum PrivacyManagerTab: String, CaseIterable {
+    /// having it empty string instructs the rendering app to use the defult tab of PM
```
andresilveirah, line 12: _or the pm tab specified in the "Show Options" action_

vilasma: Will update the comment

andresilveirah:
@vilasma I believe we should change the implementation of this feature.
Instead of setting a property on `GDPRConsentViewController` , the idea is to add a second `loadPrivacyManager` method that accepts a `PrivacyManagerTab` as parameter. 

Something like:

```swift
func loadPrivacyManager(tab: PrivacyManagerTab) {

}
```

In this way `GDPRConsentViewController` doesn't need to keep a reference to the pm tab and it can pass that information straight to the `MessageWebViewController`.

vilasma:
> @vilasma I believe we should change the implementation of this feature.
> Instead of setting a property on `GDPRConsentViewController` , the idea is to add a second `loadPrivacyManager` method that accepts a `PrivacyManagerTab` as parameter.
> 
> Something like:
> 
> ```swift
> func loadPrivacyManager(tab: PrivacyManagerTab) {
> 
> }
> ```
> 
> In this way `GDPRConsentViewController` doesn't need to keep a reference to the pm tab and it can pass that information straight to the `MessageWebViewController`.



> @vilasma I believe we should change the implementation of this feature.
> Instead of setting a property on `GDPRConsentViewController` , the idea is to add a second `loadPrivacyManager` method that accepts a `PrivacyManagerTab` as parameter.
> 
> Something like:
> 
> ```swift
> func loadPrivacyManager(tab: PrivacyManagerTab) {
> 
> }
> ```
> 
> In this way `GDPRConsentViewController` doesn't need to keep a reference to the pm tab and it can pass that information straight to the `MessageWebViewController`.

@andresilveirah  

1.  with this approach will load the requested `PM Tab` only when the user calls `loadPrivacyManager` method.
2. In the case client uses `loadMessage/loadNativeMessage` method, PM will be loaded with default tab in native message case and in web message will be loaded with tab assigned to message in analytics portal.

andresilveirah:
@vilasma 🤔 correct, I didn't think about that. Let's continue with the current implementation then.

PR #280, opened 2021-01-27
[SP-6062] Avoid sending error data to DataDog during UI tests

vilasma:


andresilveirah:
@vilasma I have 2 suggestions. We could:
1. set and use an environment variable ([example](https://stackoverflow.com/questions/37800790/hide-strange-unwanted-xcode-logs/39461256#39461256)) to decide wether to send metrics or not. or;
2. only send metrics if we know we are in a "release" environment (rather than debug).

Furthermore, instead of setting a flag, we can simply instantiate a no-op logger (that implements the same protocol) when we determine no metrics should be sent.

PR #302, opened 2021-06-01
[SP-7047] Add OS Version and Msg Id to Apple tracking API request

vilasma:


ConsentViewController/Classes/SourcePointClient/IDFAStatusReportRequest.swift
```diff
@@ -8,8 +8,9 @@
 import Foundation
 
 struct AppleTrackingPayload: Codable {
-    let appleMsgId: Int?
     let appleChoice: SPIDFAStatus
+    let appleMsgId: Int?
+    let partition_uuid: String?
```
andresilveirah, line 13: Similar case here, we can rename the attribute from `partition_uuid` to `messagePartitionUUID`. Since now the attribute will have a different name than the encoded payload, you'll have to override the `CodingKey` ([example](https://github.com/SourcePointUSA/ios-cmp-app/blob/a9eb499f615ec548bf6dc45494444990477e6aa4/ConsentViewController/Classes/SourcePointClient/ConsentRequestResponse.swift#L24))

ConsentViewController/Classes/SourcePointClient/MessageResponse.swift
```diff
@@ -88,6 +88,7 @@ struct MessageMetaData: Decodable, Equatable {
     let categoryId: MessageCategory
     let subCategoryId: MessageSubCategory
     let messageId: Int
+    let prtnUUID: String
```
andresilveirah, line 91: Same here. We can give `prtnUUID` a friendlier name and override `CodingKey` enum to match the payload name.

andresilveirah (approved):
💯

PR #313, opened 2021-06-23
Use messageLanguage property in getMessages request

mskrischke:
The messageLanguage property on SPConsentManager is now passed into the getMessages method and allows retrieval of localized consent messages on iOS.


andresilveirah:
Hi @mskrischke thank you very much for taking the time to report the issue *and* opening a PR to fix it!
I'm merging it and you can expect a release soon.

PR #320, opened 2021-07-01
add propertyId to local storage

andresilveirah:
Property id is necessary when loading a PM that belongs to a property group. 
We currently receive `propertyId` from the call to `/get_messages` and keep it in memory. But if the SDK is instantiated only for loading the PMs, the field would be `nil`. 
This PR fixes this issue by storing the `propertyId` in the local storage (UserDefaults) and using it whenever necessary.


vilasma (approved):
❤️

PR #347, opened 2021-11-08
Avoid force unwrap for nativePMMessage

alexShepard84:
We have encountered a crash due to a force unrwap under tvOS. Error is TBD


marcelosarquis (approved):

andresilveirah:
Thank you for the contribution @alexShepard84 and @marcelosarquis 🙌

PR #352, opened 2021-12-16
SP-8330 appletv html crash

Nevazhnovu:
Trying to replace html tags with attributed string was bad idea. So they are simply stripped out in order to avoid the crash


andresilveirah (commented):
@Nevazhnovu have you committed an empty file by mistake (`NSAttributedString.swift`) ?
What do you think of adding a unit test for the new method? Making sure it does remove the HTML tags from strings?

Nevazhnovu:
@andresilveirah Yup I did it accidentaly. Sure, starting right now

andresilveirah (approved):

PR #356, opened 2022-01-13
SP-8346 fix old unit tests

Nevazhnovu:
There are AuthIDUITests test suite which has 2 failing tests and PMAsFirstLayerMessageUITests test suite which has 1 failing test since there is a bug described in https://sourcepoint.atlassian.net/browse/SP-8529


Nevazhnovu:
Hmm. ObjC-ExampleApp-tests and SourcePointMetaAppUITests.Edit_Property_from_property_list are passing locally but failing here, gotta check what is going on 🤔

andresilveirah (approved):

PR #357, opened 2022-01-17
[SP-8236] "https://" is not included in SPPropertyName if it is already there

Nevazhnovu:


andresilveirah (commented):
@Nevazhnovu I believe this should not be added to the SDK. Property names should not include `https://` . We only prefix it because of legacy reasons.

Nevazhnovu:
> @Nevazhnovu I believe this should not be added to the SDK. Property names should not include `https://` . We only prefix it because of legacy reasons.

Nowadays we add `https://` even if property name already include it. Such behaviour leads to "Unexpected /message response" errors. You sure?

andresilveirah:
Let me double check with some folks. I thought property names should not include `https://` but it doesn't hurt checking.

andresilveirah:
@Nevazhnovu alright, I checked and we do have some properties with `http://|https://` in the name. I think we should extend your fix to check for `http` as well. If it has it, we should prefix the prop name with `https://`.

Nevazhnovu:
> we should prefix the prop name with `https://`.
Just to be sure I understood right: "http://prop" -> "**https**://prop", right ?

Nevazhnovu:
@andresilveirah replacing is done, pls validate

andresilveirah:
@Nevazhnovu I believe I wasn't clear. The logic should be:
1. if the prop name starts with https:// or http://, it should remain as is
2. else, we prefix it with https://

Nevazhnovu:
@andresilveirah I got you. Commit is on the way, stay tuned

Nevazhnovu:
@andresilveirah check it out

andresilveirah:
@Nevazhnovu wonderful. We just need 2 unit tests covering the changes ❤️

Nevazhnovu:
@andresilveirah check it out

andresilveirah:
💯

andresilveirah (approved):

PR #363, opened 2022-02-07
SP-8691 fix pubData onAction

andresilveirah:
All delegate method run asynchronously. Therefore the `action.pubData` property was not being updated synchronously before the consent request.
This PR makes sure the `onAction` runs on the main thread (synchronously).

### DIFF
--- a/ConsentViewController/Classes/SPConsentManager.swift
+++ b/ConsentViewController/Classes/SPConsentManager.swift
@@ -497,9 +497,7 @@ extension SPConsentManager: SPMessageUIDelegate {
     }
 
     func onAction(_ action: SPAction, from controller: UIViewController) {
-        DispatchQueue.main.async { [weak self] in
-            self?.delegate?.onAction(action, from: controller)
-        }
+        delegate?.onAction(action, from: controller)
     }
 }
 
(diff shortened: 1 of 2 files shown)


ConsentViewController/Classes/SPConsentManager.swift
```diff
@@ -497,9 +497,7 @@ extension SPConsentManager: SPMessageUIDelegate {
     }
 
     func onAction(_ action: SPAction, from controller: UIViewController) {
-        DispatchQueue.main.async { [weak self] in
```
Nevazhnovu, line 500: @andresilveirah  why is it not replaced with "DispatchQueue.main.sync"? Just curious.

andresilveirah: @Nevazhnovu that's a good question. I first tried using `main.sync` but it turns out if the process is being executed in the main thread, running `main.sync` throws a run time exception. The solution is to simply run the code you want without dispatching it to the main queue.

Nevazhnovu: @andresilveirah ahh sweet sweet deadlock. Got it

Nevazhnovu:
oh sh.. I'm sorry

Nevazhnovu (approved):

PR #367, opened 2022-03-08
SP-8864 remove default logo

Nevazhnovu:
Default logo from xibs removed + fade-in animation implemented

### DIFF
--- a/ConsentViewController/Classes/Views/tvOS/NativePrivacyManager/Common/SPNativeScreenViewController.swift
+++ b/ConsentViewController/Classes/Views/tvOS/NativePrivacyManager/Common/SPNativeScreenViewController.swift
@@ -14,7 +14,11 @@ extension UIImageView {
             if let data = try? Data(contentsOf: url) {
                 if let image = UIImage(data: data) {
                     DispatchQueue.main.async {
+                        self?.alpha = 0
                         self?.image = image
+                        UIView.animate(withDuration: 0.5, delay: 0.0, options: UIView.AnimationOptions.curveEaseIn, animations: {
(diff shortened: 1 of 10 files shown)


ConsentViewController/Classes/Views/tvOS/NativePrivacyManager/Common/SPNativeScreenViewController.swift
```diff
@@ -14,7 +14,11 @@ extension UIImageView {
             if let data = try? Data(contentsOf: url) {
                 if let image = UIImage(data: data) {
                     DispatchQueue.main.async {
+                        self?.alpha = 0
                         self?.image = image
+                        UIView.animate(withDuration: 0.5, delay: 0.0, options: UIView.AnimationOptions.curveEaseIn, animations: {
+                            self?.alpha = 1.0
+                        }, completion: nil)
```
andresilveirah, line 21: Just two Swift tips
1. when the completion handler is nil you can often omit it, for example:
```swift
UIView.animate(withDuration: 0.5, delay: 0.0, options: UIView.AnimationOptions.curveEaseIn, animations: {
  self?.alpha = 1.0
}, completion: nil)
```
can be written like:
```swift
UIView.animate(withDuration: 0.5, delay: 0.0, options: UIView.AnimationOptions.curveEaseIn, animations: {
  self?.alpha = 1.0
})
```
2. enums can most of the time be shorthanded as well, so this:
```swift
UIView.AnimationOptions.curveEaseIn
```
can be replaced by
```swift
.curveEaseIn
```

andresilveirah (approved):
💯

PR #368, opened 2022-03-11
[SP-8866] coloured subview added as underlay to tvOS button

Nevazhnovu:
Fix of tvOS button color when focused


andresilveirah (approved):
Looking good!

PR #370, opened 2022-03-22
[SP-8880] focusGuide added between acceptButton and every cell in categoryTableView + between every button in actionButtons UIStack

Nevazhnovu:
Now user cannot be stuck in the topmost cells of categoryTableView and left navigation click always leads to "Accept All" button


andresilveirah:
1. @Nevazhnovu do we have to apply the same fix for the CCPA views as well?
2. I found that when navigating left and right between the table view and buttons, whenever the focus get back to the table view it goes to the topmost element (check below).
Simulator Screen Recording - Apple TV 4K (at 1080p) (2nd generation) - 2022-03-23 at 10 48 06
Do you think we could store the position and section the focus was before moving to the button and when focusing back on the tableview the user lands on the same item?

Nevazhnovu:
@andresilveirah done!

andresilveirah (approved):

PR #373, opened 2022-03-28
SP-8966 remove purpose description

Nevazhnovu:
Purpose description label removed from first CCPA & GDPR views


andresilveirah:
@Nevazhnovu I think removing the purpose description from the 1st layer is great. It's so great that we could go one step further and remove it from the `Manage Preferences` screens (GDPR and CCPA) as well. 


GDPR_ManagePrefsScreen

PR #374, opened 2022-04-04
[SP-7484] support of `groupPmId` and `childPmId` has been added

Nevazhnovu:


ConsentViewController/Classes/SPSDK.swift
```diff
@@ -9,11 +9,13 @@ import Foundation
 
 @objc public protocol SPCCPA {
     @objc func loadCCPAPrivacyManager(withId id: String, tab: SPPrivacyManagerTab)
+    @objc func loadCCPAPrivacyManagerChildPM(tab: SPPrivacyManagerTab)
```
andresilveirah, line 12: should the signature be:
```swift
@objc func loadCCPAPrivacyManagerChildPM(withFallbackId id: String, tab: SPPrivacyManagerTab)
```
(same as GDPR?)

ConsentViewController/Classes/LocalStorage/SPUserDefaults.swift
```diff
@@ -17,6 +17,8 @@ class SPUserDefaults: SPLocalStorage {
     static let USER_DATA_KEY = "\(SP_KEY_PREFIX)userData"
```
andresilveirah, line 17: @Nevazhnovu in this class we don't need setters and getters, we can directly use properties (like the other attributes). Something like

```swift
    var gdprChildPMId: String? {
        get {
            storage.string(forKey: gdpr child PM id key)
        }
        set {
            storage.set(newValue, forKey: gdpr child PM id key)
        }
    }

    var ccpaChildPMId: String? {
        get {
            storage.string(forKey: gdpr child PM id key)
        }
        set {
            storage.set(newValue, forKey: gdpr child PM id key)
        }
    }
  ```
  
  And we can't forget to add those properties to the `SPLocalStorage` protocol.

**Review by andresilveirah — CHANGES_REQUESTED** [review]

**Review by carmelo-iriti — APPROVED** [review]

**Comment by carmelo-iriti** [comment]
@andresilveirah should we expose the `childPmId` in the `SPCCPAConsent/SPGDPRConsent` object?

**Comment by andresilveirah** [comment]
@carmelo-iriti good point. It's unfortunate we get the `childPmId` in the `userConsents` object. @Nevazhnovu the `childPmId` should be public in the consent classes.

**Comment by Nevazhnovu** [comment]
> @Nevazhnovu the `childPmId` should be public in the consent classes.

@andresilveirah It is already a part of `SPCCPAConsent` and `SPGDPRConsent`. Is it wrong?

File: ConsentViewController/Classes/SPError.swift [hunk 3]
```diff
@@ -75,6 +75,23 @@ import Foundation
     required init?(coder aDecoder: NSCoder) { fatalError("init(coder:) has not been implemented") }
 }
 
+@objcMembers public class LogCustomMetricsChildPmIdError: SPError {
```
andresilveirah, line 78: @Nevazhnovu @carmelo-iriti are we using the same `spCode` in both SDKs?

@Nevazhnovu can I suggest naming the class `MissingChildPmIdError`?

ConsentViewController/Classes/SPSDK.swift
```diff
@@ -39,6 +39,7 @@ import Foundation
         legIntCategories: [String],
         handler: @escaping (SPGDPRConsent) -> Void
     )
+    @objc func logErrorMetrics(_ error: SPError)
```
andresilveirah, line 42: @Nevazhnovu I'm not sure we should expose `logErrorMetrics` method. This method seems very internal to the SDK in my opinion.

Nevazhnovu: @andresilveirah Yes it should be internal. But 1) using `onError` will trigger clearing all data from storage which is not desirable 2) we cannot declare `private` methods in protocols. What's the best option here?

ConsentViewController/Classes/SPError.swift
Nevazhnovu: 1) 99% that code is different
2) no problem!

ConsentViewController/Classes/SPSDK.swift
Nevazhnovu: @andresilveirah I'm not used to make logs to `custom_metrics` so this entire commit may not be done in a smart way. Looking forward to see your suggestions!

ConsentViewController/Classes/LocalStorage/SPUserDefaults.swift
```diff
@@ -63,10 +66,33 @@ class SPUserDefaults: SPLocalStorage {
...
     }
 
+    var gdprChildPMId: String? {
+        get { storage.string(forKey: getChildPmIdKey(type: .gdpr)) }
+        set { storage.set(newValue, forKey: getChildPmIdKey(type: .gdpr)) }
+    }
+
+    var ccpaChildPMId: String? {
+        get { storage.string(forKey: getChildPmIdKey(type: .ccpa)) }
+        set { storage.set(newValue, forKey: getChildPmIdKey(type: .ccpa)) }
```
andresilveirah, line 76: @Nevazhnovu I believe you don't need the method `getChildPmIdKey` since now you know you're storing either CCPA or GDPR child pm id.

ConsentViewController/Classes/SPConsentManager.swift
```diff
@@ -417,6 +434,14 @@ import UIKit
             }
         }
     }
+
+    public func loadCCPAPrivacyManagerChildPM(withFallbackId id: String, tab: SPPrivacyManagerTab = .Default) {
```
andresilveirah, line 438: @Nevazhnovu should we rename this method to `loadCCPAGroupPrivacyManager(...)` ?
The developer doesn't (and shouldn't) know about the existence of a "child pm id". For our user, they are loading a group pm and not a child pm. 

Same for GDPR below.

ConsentViewController/Classes/SPSDK.swift
```diff
@@ -56,11 +57,24 @@ public extension SPSDK {
         loadMessage(forAuthId: authId, publisherData: pubData)
     }
 
-    func loadCCPAPrivacyManager(withId id: String, tab: SPPrivacyManagerTab = .Default) {
-        loadCCPAPrivacyManager(withId: id, tab: tab)
+    func loadCCPAPrivacyManager(withId id: String, tab: SPPrivacyManagerTab = .Default, useGroupPmIfAvailable: Bool = false) {
```
andresilveirah, line 60: @Nevazhnovu you first need to add:
```swift
loadCCPAPrivacyManager(withId: id, tab: SPPrivacyManagerTab, useGroupPmIfAvailable: Bool)
```
to `SPCCPA` protocol (the method is already there but it's missing the last parameter).

Then all you need to do in the extension is to call it with the default values (you don't need to add any particular concrete implementation). Eg.
```swift
func loadCCPAPrivacyManager(withId id: String, tab: SPPrivacyManagerTab = .Default, useGroupPmIfAvailable: Bool = false) {
    loadCCPAPrivacyManager(withId: id, tab: tab, useGroupPmIfAvailable: useGroupPmIfAvailable)
}
```
And do the same for GDPR.

PS.
We use the extension only because Swift Protocols don't allow for arguments with default argument. So we _extend_ it to add the method with default arguments.

andresilveirah (changes requested):

ConsentViewController/Classes/LocalStorage/SPUserDefaults.swift
Nevazhnovu: @andresilveirah do you mean we can store it under the same key? But what if we'll have both GDPR & CCPA `childPmId` included into json response

Nevazhnovu: Uh oh I got the idea. Will hardcode both keys and that's it

Nevazhnovu:
@carmelo-iriti we should sync on `spCode` for `custom_metrics`

ConsentViewController/Classes/SPConsentManager.swift
```diff
@@ -328,13 +328,14 @@ import UIKit
 
     public func loadGDPRPrivacyManager(withId id: String, tab: SPPrivacyManagerTab = .Default, useGroupPmIfAvailable: Bool = false) {
         messagesToShow += 1
-        let usedId: String
-        if useGroupPmIfAvailable, let childPmId = storage.gdprChildPMId {
-            usedId = childPmId
-        } else {
-            usedId = id
+        var usedId: String = id
```
carmelo-iriti, line 331: Just a stylist consideration: nested ifs are not so readable, you can consider to flat them in only one.

Nevazhnovu: @carmelo-iriti totally makes sense

ConsentViewController/Classes/SPConsentManager.swift
```diff
@@ -262,6 +262,19 @@ import UIKit
...
     }
+
+    private func selectPrivacyManagerId(fallbackId:String, groupPmId:String?, childPmId:String?) -> String
+    {
+        let hasGroupPmId = groupPmId != nil && groupPmId != ""
+        let hasChildPmId = childPmId != nil && childPmId != ""
+        if hasChildPmId, let childPmId = childPmId {
+            return childPmId
+        }
+        if hasGroupPmId, !hasChildPmId {
```
carmelo-iriti, line 273: Beautiful! 🤩

ConsentViewController/Classes/LocalStorage/SPUserDefaults.swift
```diff
@@ -17,6 +17,9 @@ class SPUserDefaults: SPLocalStorage {
     static let USER_DATA_KEY = "\(SP_KEY_PREFIX)userData"
     static let IAB_CMP_SDK_ID_KEY = "\(IAB_KEY_PREFIX)CmpSdkID"
     static let IAB_CMP_SDK_ID = 6
+    private let GDPR = "GDPR"
```
andresilveirah, line 20: @Nevazhnovu what about:
```swift
static let GPDR_CHILD_PM_ID_KEY = "\(SPUserDefaults.SP_KEY_PREFIX)\(GDPR)childPmId"
```
This way you can just use this value when getting/setting the value from storage, instead of "building" the key in the getter/setters.

Same is valid for the CCPA version.

ConsentViewController/Classes/LocalStorage/SPUserDefaults.swift
```diff
@@ -17,6 +17,9 @@ class SPUserDefaults: SPLocalStorage {
...
+    private let CCPA = "CCPA"
+    private let UNKNOWN = "UNKNOWN"
```
andresilveirah, line 22: I believe `UNKNOWN` is no longer used.

ConsentViewController/Classes/SPConsentManager.swift
andresilveirah: Nice touch encapsulating the logic to select the pm id in a method 💯

README.md
```diff
@@ -370,6 +370,29 @@ Utilize the following method if an end-user requests to have their data deleted:
...
+## Set a Property Group Id
+ Property groups allow your organization to group properties together in order to simplify configurations for  mass campaigns and updates. 
+ In order to use a Property Group Id, you should edit the SDK configuration object as follows:
+ 
+ ```swift
+ lazy var consentManager: SPConsentManager = { SPConsentManager(
+    accountId: 22,
+    propertyName: try! SPPropertyName("mobile.multicampaign.demo"),
+    campaigns: SPCampaigns(
+        gdpr: SPCampaign(groupPmId: "123") // <- "123" is the property group id
```
andresilveirah, line 382: ~// <- "123" is the property group id~
// <- "123" is the id of the privacy manager for the property group

andresilveirah: Same is valid for the content below. We should substitute _Property Group Id_ for _Privacy Manager Id for the Property Group_.

Nevazhnovu:
@andresilveirah Done!

andresilveirah (approved):

PR #375, opened 2022-05-05
Update 'swift.yml' to include missing test suits

Nevazhnovu:
2 unit suits added to swift.yaml


Nevazhnovu:
For whom it may concern it's just a draft PR where I've tested `swift.yaml` changes. No need to review or merge. It can be deleted as soon as #376 is reviewed

PR #376, opened 2022-05-10
[SP-8701] fix unit test suites

Nevazhnovu:


Nevazhnovu:
There are some flaky unit test which are not passing consistently. 
These two are not passing sometimes in GitHub CI but always pass locally: 
`SimpleClientSpec.request__calls_async_on_its_dispatchQueue_with_the_result_of_the_dataTask()`
`SimpleClientSpec.request__calls_dataTask_on_its_urlSession_passing_the_urlRequest_as_param()`
This one sometimes is not passing even locally due to known Nimble issue with `.toEventually(...)` (`Await.wait()` – which is Nimble internal method – throws an exception for some reason) :
`SourcePointClient.__calls_POST_on_the_http_client_with_the_right_body()`

Example/SourcePointMetaAppUITests/Helpers/MetaApp.swift
```diff
@@ -399,11 +400,11 @@ extension MetaApp: GDPRUI {
     }
 
     var PersonalisedAdsSwitch: XCUIElement {
-        consentUI.switches["Create a personalised ads profile"].firstMatch
+        consentUI.switches.element(boundBy: 2) // == "Create a personalised ads profile"
```
andresilveirah, line 403: @Nevazhnovu why did you opt for changing these lines? to make it more robust to changes on the text? If that's the case, I think switching from 
```
consentUI.switches["Create a personalised ads profile"].firstMatch
```
to 
```
consentUI.switches.element(boundBy: 2)
```
makes it less readable (though more robust). I'd still go with the more readable approach.

ConsentViewController/Classes/Consents/SPUserData.swift
```diff
@@ -54,6 +54,25 @@ extension SPConsent {
     }
 
     public override var description: String { "gdpr: \(String(describing: gdpr)), ccpa: \(String(describing: ccpa))" }
+
+    open override func isEqual(_ object: Any?) -> Bool {
```
andresilveirah, line 58: Does the parameter `object` need to be of type `Any?`? Or could you change it to `SPUserData` and remove that casting (`as?`) from the code?

Example/ConsentViewController_ExampleTests/Extensions/GDPRWebViewExtensionsSpec.swift
```diff
@@ -29,11 +29,16 @@ class TestWebView: WKWebView, WKNavigationDelegate {
 
 class GDPRWebViewExtensionsSpec: QuickSpec {
     override func spec() {
+        beforeSuite {
+            // changing AsyncDefaults make the test suite pass in CI due to slow CI environment
+            AsyncDefaults.timeout = .seconds(5)
```
andresilveirah, line 34: we might have to increase it to 20s sometimes...

Example/ConsentViewController_ExampleTests/Extensions/GDPRWebViewExtensionsSpec.swift
```diff
@@ -44,24 +49,25 @@ class GDPRWebViewExtensionsSpec: QuickSpec {
                     webview.getAuthId { result, _ in authId = result }
                 }
                 webview.loadHTMLString("", baseURL: URL(string: "https://sourcepoint.com")!)
-
                 expect(authId).toEventually(equal("foo"))
             }
         }
 
         describe("setAuthId") {
+            beforeSuite {
```
andresilveirah, line 57: Since you already changed the timeout in the top most `beforeSuite` you don't need to do it again in the same file.

Example/SourcePointMetaAppUITests/Helpers/MetaApp.swift
Nevazhnovu: @andresilveirah it somehow doesn't detect the element if search by it's name but works perfectly fine with index

ConsentViewController/Classes/Consents/SPUserData.swift
Nevazhnovu: @andresilveirah ` override func isEqual(_ object: SPUserData) ` goes `Method does not override any method from its superclass`
since `NSObject`'s method declaration is ` override func isEqual(_ object: Any?) `
I can: 
1) get rid of `override` by using overloading ( ` func isEqual(_ object: SPUserData) ` ) 
or 
2) override the ` func ==(left: SPUserData, right: SPUserData) -> Bool `

Example/ConsentViewController_ExampleTests/Extensions/GDPRWebViewExtensionsSpec.swift
Nevazhnovu: @andresilveirah does this change affects only this suite or the whole set of suites it belongs to? Do I need to set it back to default value after this suite is done?

Example/ConsentViewController_ExampleTests/Extensions/GDPRWebViewExtensionsSpec.swift
Nevazhnovu: Done

Example/SourcePointMetaAppUITests/Helpers/MetaApp.swift
andresilveirah: Hm... and the text is definitely definitely there?

ConsentViewController/Classes/Consents/SPUserData.swift
andresilveirah: Oh, my bad, just saw the standard implementation of all `func isEqual` takes `Any?` as param. Let's stick with it.

carmelo-iriti (approved):

Example/ConsentViewController_ExampleTests/Extensions/GDPRWebViewExtensionsSpec.swift
andresilveirah: That's a good question @Nevazhnovu . According to [Quick's documentation](http://quick.github.io/Quick/Functions.html#/s:5Quick11beforeSuiteyyyycF), that block of code will run only once per suite of specs (e.g. per file). However, since [`AsyncDefaults`](https://quick.github.io/Nimble/Structs/AsyncDefaults.html#/s:6Nimble13AsyncDefaultsV7timeout8Dispatch0E12TimeIntervalOvpZ) is a global, I'd say it's of good practice to add a:
```swift
afterSuite {
    AsyncDefaults.timeout = _default value_
}
```

Example/SourcePointMetaAppUITests/Helpers/MetaApp.swift
Nevazhnovu: @andresilveirah  120% sure

Example/ConsentViewController_ExampleTests/Extensions/GDPRWebViewExtensionsSpec.swift
Nevazhnovu: Done

andresilveirah (approved):

PR #378, opened 2022-07-22
[DIA-1000] gracefully degradation onError

Nevazhnovu:
Now `SPConsentManager` fires `onConsentReady` instead of `onError` IF ` SPUserData` is not empty in `SPUserDefaults`

### DIFF
--- a/ConsentViewController/Classes/SPConsentManager.swift
+++ b/ConsentViewController/Classes/SPConsentManager.swift
@@ -249,9 +249,7 @@ import UIKit
         let userData = storage.userData
         if !userData.isEqual(SPUserData()) {
             delegate?.onConsentReady?(userData: userData)
-        }
-        else
-        {
+        } else {
             if cleanUserDataOnError {
                 SPConsentManager.clearAllData()
             }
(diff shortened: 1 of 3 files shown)


andresilveirah:
@Nevazhnovu we should only "graceful degrade" when calling `loadMessage`. In your implementation, all calls to `ConsentManager.onError` will be "gracefully degraded". 

For example, if there's an error after the user taps on Accept all (e.g. during the call to `consent` endpoint), that error should bubble up to the delegate and not be gracefully handled.

Nevazhnovu:
@andresilveirah gotcha, assign me on the task, I will fix it.

ConsentViewController/Classes/SPConsentManager.swift
```diff
@@ -17,7 +17,7 @@ import UIKit
     /// By default, the SDK will remove all user consent data from UserDefaults, possibly triggering a message to be displayed again next time
     /// `.loadMessage` is called.
     /// Set this flag to `false` if you wish to opt-out from this behaviour.
```
andresilveirah, line 19: @Nevazhnovu can you update the comment to reflect the new behaviour?

Nevazhnovu: @andresilveirah fixed, as well as `README`

Nevazhnovu:
@andresilveirah pls, review or approve

PR #379, opened 2022-08-08
Add deep link to FAQs

tom-win87:


tom-win87:
@andresilveirah Added custom action solution to deep linking

andresilveirah:
Thank you @tom-win87 !

PR #382, opened 2022-09-07
DIA-1180 implement call to messages

andresilveirah:


carmelo-iriti (approved):
The only thing that I'm a bit unsure is the nullability on some fields like `tcfData`, etc.. But this is out of our control right now

andresilveirah:
> The only thing that I'm a bit unsure is the nullability on some fields like `tcfData`, etc.. But this is out of our control right now

Agree with you @carmelo-iriti . I actually went back and forth making `TCData` optional... We can still change our minds before releasing `v7` if it makes sense.

PR #398, opened 2022-12-05
[HCD-327] Google Additional Consent

tom-win87:
Google additional consent string is stored in `IABTCF_AddtlConsent` in user local storage


andresilveirah:
Thank you for your contribution @tom-win87

PR #413, opened 2023-01-25
DIA-1694 v7 expand ccpa status support

andresilveirah:


ConsentViewController/Classes/Consents/SPCCPAConsent.swift
```diff
@@ -24,35 +24,41 @@ import Foundation
     /// consent to all vendors and purposes.
     case ConsentedAll
 
+    case LinkedNoAction
```
andresilveirah, line 27: @carmelo-iriti it seems the CCPA consent service (used by wrapper) ca now return a new value called `"linkedNoAction"`. On iOS we used to throw an error if the consent status had an unknown value, but that's obviously a bad idea. 

With this PR, we add the new value to the list of values (enums) the SDK knows and we add a new value called `.Unknown` for the case we get a value we don't know how to parse. 

Do you do something similar on Android as well?

carmelo-iriti: on android if we receive something that we don’t recognise, we set by default to `consentedAll`

carmelo-iriti: no exception gets thrown

carmelo-iriti: Probably a better option could be using as default `rejectedNone`

PR #422, opened 2023-03-27
[DIA-1896] processor directives added for `v7` to avoid malfunction of SPM library import

Nevazhnovu:


Nevazhnovu:
@andresilveirah is this pr still needed?

PR #431, opened 2023-04-26
[DIA-2101] all the `URL`s are updated to include `scriptType` & `scriptVersion` query parameters

Nevazhnovu:


Nevazhnovu:
@andresilveirah pls, notice that I've changed the `CCPA_PM_URL` and `GDPR_PM_URL`. Are they supposed to be affected by this change?

andresilveirah:
@Nevazhnovu please have a look a the unit tests, I believe they failed because we had expectations on the URLs we use.

andresilveirah:
> @andresilveirah pls, notice that I've changed the `CCPA_PM_URL` and `GDPR_PM_URL`. Are they supposed to be affected by this change?

Good catch. No, PM urls don't need to include those new query params.

andresilveirah (approved):

PR #434, opened 2023-04-28
[DIA-1831] HTML tags for `Description` text on `HomeView` are applied properly

Nevazhnovu:


ConsentViewController/Classes/Views/tvOS/NativePrivacyManager/GDPR/SPGDPRPartnersViewController.swift
```diff
@@ -97,7 +97,7 @@ extension SPGDPRPartnersViewController: UITableViewDataSource, UITableViewDelega
 
     func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
         let label = UILabel(frame: CGRect.init(x: 0, y: 0, width: tableView.frame.width, height: 50))
-        label.text = "\(sections[section]?.settings.text.stripOutHtml() ?? "Partners")"
+        label.text = "\(sections[section]?.settings.text.stripOutHtml(noHTML: true) ?? "Partners")"
```
andresilveirah, line 100: @Nevazhnovu on a method called `stripOutHtml` why should we have a flag called `noHTML`? Shouldn't it _always_ remove html?

Nevazhnovu: @andresilveirah hmm you're right. Currently, the method strips CSS tags but allows the HTML-tagged text to be processed as `attributedText`. I wanted to make HTML stripping optional. I will rename the function to `stripOutCss`

PR #442, opened 2023-05-16
[DIA-2137] update error codes

Nevazhnovu:


ConsentViewController/Classes/SPError.swift
```diff
@@ -135,7 +136,7 @@ import Foundation
 }
 
 @objcMembers public class RenderingAppError: SPError {
-    override public var spCode: String { renderingAppErrorCode ?? "sp_metric_rendering_app_error" }
+    override public var spCode: String { renderingAppErrorCode ?? "sp_metric_webview_creation_error" }
```
Nevazhnovu, line 139: Warning! This might be an unnecessary change. We have to discuss it.

Example/ConsentViewController_ExampleTests/SPErrorSpec.swift
```diff
@@ -116,7 +116,7 @@ class SPErrorSpec: QuickSpec {
             describe("RenderingAppError") {
                 describe("if not code is provided") {
                     it("its spCode should be rendering_app_error") {
-                        expect(RenderingAppError(campaignType: .gdpr, nil).spCode) == "sp_metric_rendering_app_error"
+                        expect(RenderingAppError(campaignType: .gdpr, nil).spCode) == "sp_metric_webview_creation_error"
```
Nevazhnovu, line 119: Warning! This might be an unnecessary change. We have to discuss it.

ConsentViewController/Classes/SPError.swift
```diff
@@ -69,6 +69,7 @@ import Foundation
 }
 
 @objcMembers public class InvalidArgumentError: SPError {
+    override public var spCode: String { "sp_metric_invalid_local_data" }
```
andresilveirah, line 72: @Nevazhnovu judging by the name of the class (`InvalidArgumentError`), are you sure that's the correct `spCode`?

ConsentViewController/Classes/SPError.swift
Nevazhnovu: Gotta revert it back, one moment

Example/ConsentViewController_ExampleTests/SPErrorSpec.swift
Nevazhnovu: Gotta revert it back, one moment

ConsentViewController/Classes/SPError.swift
andresilveirah: I believe that's not correct. `RenderingAppError` is used for the case when the rendering app emits a `error` event and we capture that via the JSReceiver.

ConsentViewController/Classes/SPError.swift
```diff
@@ -257,20 +258,19 @@ import Foundation
...
+    }
+    @available(*, unavailable)
+    required init?(coder aDecoder: NSCoder) { fatalError("init(coder:) has not been implemented") }
 }
 
-@objcMembers public class PostingConsentWithoutConsentUUID: InvalidRequestError {
+@objcMembers public class PostingConsentWithoutConsentUUID: SPError {
+    override public var spCode: String { "sp_metric_invalid_consent_UUID" }
     override public var description: String {
         "Tried to post consent but the stored consentUUID is empty or nil. Make sure to call .loadMessage or .loadPrivacyManager first."
```
andresilveirah, line 275: I believe we need to change the error description. We no longer have `loadPrivacyManager` method but `loadGDPRPrivacyManager` or `loadCCPAPrivacyManager`.

ConsentViewController/Classes/SPError.swift
```diff
@@ -315,6 +315,26 @@ import Foundation
     override public var spCode: String { "sp_metric_error_converting_consent_snapshot_to_json" }
 }
 
-@objcMembers public class InvalidJSONEncodeResult: SPError {
-    override public var spCode: String { "sp_metric_error_invalid_JSON_encode_result" }
+public class NetworkCallErrorsCode {
```
andresilveirah, line 318: Could we use the `enum InvalidResponsAPICode` only and get rid of the class?

I believe the following should work:
```swift
enum InvalidResponsAPI: String {
    case META_DATA = "_meta-data"
    case CONSENT_STATUS = "_consent-status"
    case PV_DATA = "_pv-data"
    case MESSAGES = "_messages"
    case EMPTY = ""

    var code: String {
        "sp_metric_invalid_response_api\(rawValue)"
    }
}
```

Nevazhnovu:
@andresilveirah please review.
@bohdan-go-wombat you can do it too if you want, it is optional. Here is changes of SP error codes, the same error codes exist in Android as well.

ConsentViewController/Classes/SourcePointClient/SimpleClient.swift
```diff
@@ -95,28 +95,23 @@ class SimpleClient: HttpClient {
...
-                    handler(.failure(spError))
                 } else {
                     handler(.success(data))
                 }
             }
         }.resume()
     }
 
-    func post(urlString: String, body: Data?, handler: @escaping ResponseHandler) {
+    func post(urlString: String, body: Data?, apiCode: NetworkCallErrorsCode.InvalidResponsAPICode?, handler: @escaping ResponseHandler) {
```
andresilveirah, line 114: Here you could use a default parameter so you don't have to check for null (and `?? .EMPTY`). Something like:
```swift
func post(urlString: String, body: Data?, apiCode: NetworkCallErrorsCode.InvalidResponsAPICode? = .EMPTY, handler: @escaping ResponseHandler) {
```

ConsentViewController/Classes/SPError.swift
Nevazhnovu: @andresilveirah 146% correct. Judging by the use case - it should be the same as in Android.
"""PropertyName can only include letters, numbers, '.', ':', '-' and '/'. \(string) 
passed is invalid"""

ConsentViewController/Classes/SourcePointClient/SimpleClient.swift
Nevazhnovu: image

Nevazhnovu: I can overload a method if you want

ConsentViewController/Classes/SourcePointClient/SimpleClient.swift
```diff
@@ -95,28 +95,23 @@ class SimpleClient: HttpClient {
...
                 if error != nil {
-                    var spError = GenericNetworkError(request: urlRequest, response: response as? HTTPURLResponse)
                     if let response = response as? HTTPURLResponse {
                         switch response.statusCode {
-                        case 400...499:
-                            spError = ResourceNotFoundError(request: urlRequest, response: response)
-
-                        case 500...599:
-                            spError = InternalServerError(request: urlRequest, response: response)
+                        case 408:
```
Nevazhnovu, line 100: @andresilveirah attention!

Nevazhnovu: @andresilveirah I can make this error be thrown disregard of 408, just every time "something went wrong"

ConsentViewController/Classes/SPError.swift
andresilveirah: Hmm, so it seems to be we need a better error code and a better error class if that's the only time that error thrown. Something in the lines of `InvalidPropertyNameError` (code: `sp_metric_invalid_property_name`).

ConsentViewController/Classes/SourcePointClient/SimpleClient.swift
andresilveirah: @Nevazhnovu makes sense. So we have 3 possible errors being returned here:
1. InvalidResponse
2. Timeout
3. Something wrong

Correct?

ConsentViewController/Classes/SourcePointClient/SourcePointClient.swift
```diff
@@ -204,7 +204,7 @@ class SourcePointClient: SourcePointProtocol {
             "messageId": messageId,
             "includeData": "{\"categories\": {\"type\": \"RecordString\"}}"
         ])! // swiftlint:disable:this force_unwrapping
-        client.get(urlString: url.absoluteString) { result in
+        client.get(urlString: url.absoluteString, apiCode: nil) { result in
```
andresilveirah, line 207: @Nevazhnovu we can add apiCodes for these endpoints with `apiCode: nil`, feel free to pick a name you think is most descriptive and I'll review them.

Example/ConsentViewController_ExampleTests/SPClientCoordinator/SourcePointClient/SimpleClientSpec.swift
```diff
@@ -95,20 +95,21 @@ class SimpleClientSpec: QuickSpec {
...
                 expect(session.configuration.requestCachePolicy) == .reloadIgnoringLocalCacheData
             }
 
             it("calls async on its dispatchQueue with the result of the dataTask") {
+                let session = URLSessionMock()
                 let queue = DispatchQueueMock()
                 let client = SimpleClient(
                     connectivityManager: ConnectivityMock(connected: true),
                     logger: SPLoggerMock(),
-                    urlSession: URLSession.shared,
```
andresilveirah, line 107: Nice catch.

ConsentViewController/Classes/SourcePointClient/SimpleClient.swift
Nevazhnovu: @andresilveirah that's correct. 
My main concern is:
Should I erase the `switch-case` and `GenericNetworkError` and substitute them with the `ConnectionTimeoutAPIError` throw?
In Android, `TimeoutException` "is thrown to indicate that an input or output data transfer has been terminated because the thread performing it was interrupted" which looks like a generic case, not an exact 408 Server Timeout Error.

ConsentViewController/Classes/SPError.swift
Nevazhnovu: @andresilveirah this change was reverted back so the code remains untouched. My bad.

ConsentViewController/Classes/SPError.swift
Nevazhnovu: @andresilveirah changed to `sp_metric_invalid_property_name` but consider my previous message (it will not be unified with Android SDK this way)

ConsentViewController/Classes/SourcePointClient/SourcePointClient.swift
Nevazhnovu: @andresilveirah done

ConsentViewController/Classes/SPError.swift
andresilveirah: Sure, it makes sense.

ConsentViewController/Classes/SourcePointClient/SimpleClient.swift
andresilveirah: @Nevazhnovu protocols don't accept default parameters, but their concrete implementations do.

Nevazhnovu:
@andresilveirah done

PR #444, opened 2023-05-24
[DIA-2061] `childPmId` is saved to `UserDefaults` when it comes in `/meta-data` response

Nevazhnovu:
- [x] `childPmId` no longer comes from messages
- [x]  save `childPmId` which comes from `/meta-data` to `UserDefaults`
- [x] resend `childPmId`
- [x] ...
- [x] Profit!

### DIFF
--- a/Example/ConsentViewController_ExampleTests/SPClientCoordinator/SPClientCoordinatorSpec.swift
+++ b/Example/ConsentViewController_ExampleTests/SPClientCoordinator/SPClientCoordinatorSpec.swift
@@ -44,6 +44,39 @@ class SPClientCoordinatorSpec: QuickSpec {
 
         describe("a property with GDPR and CCPA campaigns") {
             describe("loadMessage") {
+                it("should return gpoupPmId from metaData and save") {
+                    coordinator = SourcepointClientCoordinator(
+                        accountId: 22,
+                        propertyName: try! SPPropertyName("mobile.prop-1"),
+                        propertyId: 24188,
+                        campaigns: SPCampaigns(
+                            gdpr: SPCampaign(groupPmId: "613056"),
+                            ccpa: SPCampaign()
+                        ),
+                        storage: LocalStorageMock()
+                    )
+                    waitUntil { done in
+                        coordinator.loadMessages(forAuthId: nil) { result in
+                            switch result {
+                            case .success(let (_, consents)):
+                                    expect(consents.gdpr?.consents?.applies).to(beTrue())
+                                    expect(consents.gdpr?.consents?.euconsent).notTo(beEmpty())
+                                    expect(consents.gdpr?.consents?.vendorGrants).notTo(beEmpty())
+                                    expect(consents.ccpa?.consents?.uspstring) == "1---"
+                                case .failure(let error):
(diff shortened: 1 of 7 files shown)


Nevazhnovu:
@andresilveirah this PR is ready for review!

andresilveirah:
@Nevazhnovu I believe this PR doesn't contain enough tests to cover this feature. Here you're only testing that `/meta-data` echoes groupPmId into its response. In the end, we're interesting in knowing the feature works as a whole, not only on that particular request.

Example/ConsentViewController_ExampleTests/SPClientCoordinator/SPClientCoordinatorSpec.swift
```diff
@@ -41,6 +41,49 @@ class SPClientCoordinatorSpec: QuickSpec {
...
+                expect(manager.selectPrivacyManagerId(fallbackId: "1", groupPmId: "2", childPmId: "3"))=="3"
+                expect(manager.selectPrivacyManagerId(fallbackId: "1", groupPmId: nil, childPmId: "3"))=="1"
+                expect(manager.selectPrivacyManagerId(fallbackId: "1", groupPmId: "2", childPmId: nil))=="1"
+                expect(manager.selectPrivacyManagerId(fallbackId: "1", groupPmId: nil, childPmId: nil))=="1"
+            }
+        }
+
+        describe("a property with GDPR and CCPA campaigns") {
+            describe("loadMessage") {
+                it("should return gpoupPmId from metaData and save") {
```
andresilveirah, line 57: @Nevazhnovu I believe you duplicated the `describe` in this file. Instead, you can move the `it("should return...` into the existing `describe("loadMessage")` a few lines below.

Example/ConsentViewController_ExampleTests/SPClientCoordinator/SourcePointClient/UnmockedSourcepointClientSpec.swift
```diff
@@ -196,6 +196,28 @@ class UnmockedSourcepointClientSpec: QuickSpec {
             }
         }
 
+        describe("meta-data_groupPmId") {
+            it("Check if groupPmId echo") {
```
andresilveirah, line 200: Same goes here, since we already have a `describe("meta-data")` you don't need to duplicate it, you can simply add the new `it("...` to the existing `describe` a few lines above.

Example/ConsentViewController_ExampleTests/SPClientCoordinator/SPClientCoordinatorSpec.swift
```diff
@@ -42,8 +42,45 @@ class SPClientCoordinatorSpec: QuickSpec {
             )
         }
 
+        describe("selectPrivacyManagerId") {
```
andresilveirah, line 45: @Nevazhnovu I believe this test doesn't belong to `SPClientCoordinatorSpec`, but rather `SPConsentManagerSpec` (feel free to create the file if that doesn't exist yet).

andresilveirah (approved):

PR #454, opened 2023-07-06
[DIA-2370] DNS button fix + USP string correction

Nevazhnovu:


Example/ConsentViewController_ExampleTests/CCPAConsentStatusSpec.swift
```diff
@@ -42,5 +42,18 @@ class CCPAConsentStatusSpec: QuickSpec {
         it("unknown values should decode to .Unknown") {
             expect("\"foo\"").to(decodeToValue(CCPAConsentStatus.Unknown))
         }
+
+        it("UPS string check") {
```
andresilveirah, line 46: It should be `USP` instead of `UPS`.

andresilveirah (approved):

Example/ConsentViewController_ExampleTests/CCPAConsentStatusSpec.swift
Nevazhnovu: Oops 😅

PR #466, opened 2023-07-24
Add markdown support to tvOS

andresilveirah:
Changes this:
Screenshot 2023-07-24 at 21 16 50

To this:
Screenshot 2023-07-24 at 21 15 39


andresilveirah:
@Nevazhnovu don't let the number of files intimidate you. We have a lot of new files because of a new dependency `Down` to handle Markdown on tvOS.

Example/Podfile.lock
```diff
@@ -1,5 +1,7 @@
 PODS:
-  - ConsentViewController (7.2.1)
+  - ConsentViewController (7.2.1):
+    - Down (~> 0.9.0)
+  - Down (0.9.5)
```
Nevazhnovu, line 4: @andresilveirah is it alright that we have two `Down` dependencies but with different versions in the Podfile?

Nevazhnovu (approved):

Example/Podfile.lock
andresilveirah: @Nevazhnovu yes. 
The notation `~> X.Y.Z` means _use the latest version as long as X and Y are the ones specified_ (so Z can be as high as possible. 
If you specify `~> X.Y` it'd mean, _use the latest version as long as X is the one specified and Y can go as high as possible. 

https://guides.cocoapods.org/syntax/podfile.html#pod

PR #470, opened 2023-07-26
Fix date formatter failing

jtyrie:
## Why:
- We observed the date formatter returning nil unexpectedly on certain devices.
- This led to the app always using the default date (now()). This means that the customers would be presented the cmp prompt every time as the last updated time is always older than the vendor list change.

### This change addresses the need by:
- Use ISO8601DateFormatter for iOS 11 and above.

### DIFF
--- a/ConsentViewController/Classes/SPDateCreated.swift
+++ b/ConsentViewController/Classes/SPDateCreated.swift
@@ -8,10 +8,16 @@
 import Foundation
 
 public struct SPDateCreated: Codable, Equatable {
-    static let format: DateFormatter = {
-        let formatter = DateFormatter()
-        formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
-        return formatter
+    static let format: SPDateFormatter = {
+        if #available(iOS 11.0, *) {
+            let formatter = ISO8601DateFormatter()
+            formatter.formatOptions = [ .withFullDate, .withFullTime, .withFractionalSeconds, .withTimeZone ]
+            return formatter
+        } else {
+            let formatter = DateFormatter()
+            formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
+            return formatter
+        }
     }()
 
     let originalDateString: String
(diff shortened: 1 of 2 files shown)


andresilveirah:
Hi @jtyrie thank you for chasing this issue and offering a fix. Could you remove the `print("JAKE ...`? I believe those were left there for your own debugging, right? 

Also, can you give me a bit more info on the devices and iOS versions you notice the issue? I'd like to try and reproduce the issue myself (have a failing test), and seeing it passing with the proposed solution.
I've tried using `ISO8601DateFormatter`, but it would not parse the date the exact way we expect. For example:

Screenshot 2023-07-26 at 15 09 17

Now I'm not sure if the test itself is wrong or the parsing of the date.

jtyrie:
Hi @andresilveirah I've removed those left over prints

Where is that test? I can't see it in the codebase 

The phone sees the issue every time, regardless of deleting apps / reinstalling etc.
It's an iPhone XS on 16.5.1(c)

This does not happen for us on most devices, but at least one other user has had the same experience, i don't know there device details though. 

The tests under `SPDateCreatedSpec` pass for me on an iPhone 14 Pro iOS 16.4 simulator

andresilveirah:
Hi @jtyrie sorry it took me some time to get back to you. I've being trying to reproduce this issue in a consistent way, but no luck. 

I just checked this StackOverflow question [Date formatter doesn't work on certain iOS Devices](https://stackoverflow.com/questions/49373149/date-formatter-doesnt-work-on-certain-ios-devices) and it seems some devices have issues with the locale. 

Do you mind adding the following code to `SPDateCreated` and seeing if it fixes the issue for the device which you can reproduce it?
```swift
public struct SPDateCreated: Codable, Equatable {
    static let format: DateFormatter = {
        let formatter = DateFormatter()
        formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
        formatter.timeZone = TimeZone(abbreviation: "GMT") // this line is new
        formatter.locale = Locale(identifier: "en_US_POSIX") // this line is new
        return formatter
    }()
    // ...
```

andresilveirah:
@jtyrie did you have the chance to use the code above on the device you were getting the issue?

jtyrie:
Hi @andresilveirah apologies, I've been on annual leave. Where did we get to with this? Do you still need me to verify?

jtyrie:
That seems to have fixed it @andresilveirah 👍 
Any idea when the new release will be ready? Cheers

andresilveirah:
Fantastic! Thank you for checking @jtyrie . The release will be ready latest by tomorrow morning. It should be `7.3.0`.

PR #478, opened 2023-08-15
Prepare release 7.3.0

andresilveirah:


tom-win87 (commented):
@andresilveirah I rewrote the section with the necessary changes. Please confirm the possible values for each attribute is correct

PR #490, opened 2023-09-11
DIA-2439 making special purposes selectable

Nevazhnovu:


andresilveirah:
@Nevazhnovu I see now special purposes are selectable. But what happens if the user selects a special purpose?

andresilveirah (approved):

PR #495, opened 2023-09-22
DIA-2389 disappear toggles if no purposes

Nevazhnovu:


carmelo-iriti (approved):

ConsentViewController/Classes/Views/tvOS/NativePrivacyManager/GDPR/SPGDPRManagePreferenceViewController.swift
```diff
@@ -137,15 +160,16 @@ extension SPGDPRManagePreferenceViewController: UITableViewDataSource, UITableVi
     }
 
     func numberOfSections(in tableView: UITableView) -> Int {
-        let legIntCategoryCount = sections.filter { $0.contentLegIntCategory.isNotEmpty() }.count
-        return displayingLegIntCategories ? legIntCategoryCount : sections.count
+        4
```
bohdan-go-wombat, line 163: is it okay, that the result of this function is hardcoded and not calculated?

Nevazhnovu: @bohdan-go-wombat Yes it is.
Previously, the UI table view made sections depending on the tab. So leg int tab might have lacked sections after showing the consent tab first. It's better to have 4 as a rule and then hide/show the sections based on the tab present.

PR #515, opened 2023-11-07
Fix UUID query param for PM URL

andresilveirah:
Fix a regression introduced by #499 in which the gdpr consent uuid value and query param would be used also for ccpa campaigns when loading the PM from the 1st layer message.

### DIFF
--- a/ConsentViewController/Classes/SPMessageUIDelegate.swift
+++ b/ConsentViewController/Classes/SPMessageUIDelegate.swift
@@ -1,5 +1,5 @@
 //
-//  MessageUIDelegate.swift
+//  MessageUIDelegateSpy.swift
 //  ConsentViewController
 //
 //  Created by Andre Herculano on 10.02.21.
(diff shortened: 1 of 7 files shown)


ConsentViewController/Classes/SPMessageUIDelegate.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  MessageUIDelegate.swift
+//  MessageUIDelegateSpy.swift
```
Nevazhnovu, line 2: @andresilveirah you haven't renamed the class or a filename, this might be a redundant change

Nevazhnovu: I see `MessageUIDelegateSpy: SPMessageUIDelegate` usage, in the sake of consistency I suggest renaming public protocol `SPMessageUIDelegate` or filename

andresilveirah: You have a point @Nevazhnovu , I'll just revert that comment back to `SPMessageUIDelegate`

PR #517, opened 2023-11-09
[DIA-2937] special feature toggle tvos

Nevazhnovu:


Example/ConsentViewController_ExampleTests/Helpers/SourcePointClientMock.swift
```diff
@@ -165,7 +165,8 @@ class SourcePointClientMock: SourcePointProtocol {
                 legIntCategories: nil,
                 legIntVendors: nil,
                 vendors: nil,
-                categories: nil
+                categories: nil,
+                specialFeatures: nil
```
bohdan-go-wombat, line 169: for the future references, can you add `,` at the end of this line, so in the future if you add something new to the constructor of `GDPRChoiceResponse`, the GH checks will not mark it as an update, only as an addition?

bohdan-go-wombat (approved):
LGTM, but don't take my word for granted here, I'm not that good at swift 😄

carmelo-iriti (approved):

Example/ConsentViewController_ExampleTests/Helpers/SourcePointClientMock.swift
andresilveirah: @bohdan-go-wombat the linter complains about dangling commas :)

ConsentViewController/Classes/Views/tvOS/NativePrivacyManager/GDPR/GDPRPMConsentSnapshot.swift
```diff
@@ -154,6 +165,13 @@ class GDPRPMConsentSnaptshot: NSObject, ConsentSnapshot, PMVendorManager, PMCate
                     legInt: toggledLIVendorsIds.contains(id),
                     vendorType: vendor.vendorType
                 )
+            },
+            specialFeatures: toggledSpecialFeatures.compactMap { id in
+                guard let feature = specialFeatures.first(where: { f in f._id == id}) else { return nil }
+                return GDPRPMPayload.Feature(
+                    _id: id,
+                    iabId: feature.iabId
+                )
```
andresilveirah, line 174: @Nevazhnovu can you use the short notation for the `.first` method?
```swift
specialFeatures.first { feature in feature._id == id } else { return nil }
```
or
```swift
specialFeatures.first { $0._id == id } else { return nil }
```

ConsentViewController/Classes/Views/tvOS/NativePrivacyManager/GDPR/GDPRPMConsentSnapshot.swift
```diff
@@ -11,6 +11,12 @@ protocol ConsentSnapshot {
     func toPayload(language: SPMessageLanguage, pmId: String) -> JSONAble
 }
 
+public enum CategoryContentType: Int {
+    case consent = 1
+    case legitimate = 2
+    case specialFeatures = 3
+}
```
andresilveirah, line 18: I like that you created an enum for that. Does it need to inherit from `Int`?

Example/ConsentViewController_ExampleTests/Helpers/SourcePointClientMock.swift
Nevazhnovu: @bohdan-go-wombat Yeah, the swift language does not support commas after the last parameter of the declaration

PR #519, opened 2023-11-10
DIA-2989 handle pm for usnat

andresilveirah:
* handle usnat actions on `JSReceiver` (to load usnat PM via 1st layer message)
* implemented `loadUSNatPrivacyManager`
* refactored example app to include usnat pm


ConsentViewController/Assets/javascript/SPJSReceiver.js
```diff
@@ -42,7 +42,7 @@ var getActionFromMessage = function (eventData) {
     return {
         id: String(choiceData.choice_id),
         type: choiceData.type,
-        pm_url: choiceData.iframe_url,
+        pmUrl: choiceData.pmUrl || choiceData.iframe_url,
```
carmelo-iriti, line 45: is this a new url to add different from before?

carmelo-iriti: ok, I saw below and I got the meaning

carmelo-iriti (approved):

ConsentViewController/Assets/javascript/SPJSReceiver.js
andresilveirah: just renaming the field :)

PR #529, opened 2023-12-11
DIA-3249 handle ccpa opt-outs

andresilveirah:
* fixed an issue with `/pv-data` being called for campaigns that are not in the config ✅ 
* fixed an issue with `/consent-status` being called in fresh installs (when `localVersion` is `nil`) ✅ 
* added `transitionCCPAAuth` flag to `SPConfig` so clients can choose to transfer their CCPA authenticated consent to USNat ✅
* implemented logic to transfer CCPA opted-out consent over to USNat ✅

### DIFF
--- a/ConsentViewController/Classes/SourcePointClient/ConsentStatusMetadata.swift
+++ b/ConsentViewController/Classes/SourcePointClient/ConsentStatusMetadata.swift
@@ -13,9 +13,41 @@ struct ConsentStatusMetaData: QueryParamEncodable {
         let applies: Bool
         let dateCreated: SPDate?
         let uuid: String?
-        let hasLocalData: Bool
+        let hasLocalData = false
         let idfaStatus: SPIDFAStatus?
     }
 
-    let gdpr, ccpa, usnat: Campaign?
+    struct USNatCampaign: Encodable {
+        let applies: Bool
+        let dateCreated: SPDate?
+        let uuid: String?
+        let hasLocalData = false
+        let idfaStatus: SPIDFAStatus?
+    }
+
+    let gdpr, ccpa: Campaign?
+    let usnat: USNatCampaign?
+}
+
+extension ConsentStatusMetaData.Campaign {
+    init?(_ consent: CampaignConsent?, campaign: SPCampaign?, idfaStatus: SPIDFAStatus?) {
+        guard let consent = consent, campaign != nil else { return nil }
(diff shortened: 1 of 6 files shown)


ConsentViewController/Classes/SourcePointClient/ConsentStatusMetadata.swift
```diff
@@ -13,9 +13,48 @@ struct ConsentStatusMetaData: QueryParamEncodable {
         let applies: Bool
         let dateCreated: SPDate?
         let uuid: String?
-        let hasLocalData: Bool
+        let hasLocalData = false
```
carmelo-iriti, line 16: do we need to set a default?

ConsentViewController/Classes/SourcePointClient/SourcepointClientCoordinator.swift
```diff
@@ -698,21 +716,21 @@ class SourcepointClientCoordinator: SPClientCoordinator {
 
     func pvData(pubData: SPPublisherData?, _ handler: @escaping () -> Void) {
         let pvDataGroup = DispatchGroup()
-        if let gdprMetadata = state.gdprMetaData {
+        if let gdprMetadata = state.gdprMetaData, campaigns.gdpr != nil {
```
carmelo-iriti, line 719: Is this to filter campaign?

ConsentViewController/Classes/SourcePointClient/ConsentStatusMetadata.swift
andresilveirah: It was always hardcoded `false`, but in another point of the code. In this line, I hardcode directly in the class, so other parts of the code don't have to set it to `false`.

ConsentViewController/Classes/SourcePointClient/SourcepointClientCoordinator.swift
andresilveirah: Yes. To give you more context, the SDK stores an object called `gdprMetaData` in the response of `/meta-data`. But we never clear those objects. So, if in the past, the got a response for GDPR, but the config no longer has a GDPR campaign, we would still have a `gdprMetaData` object stored. 

This change basically checks if we have the `gdprMetaData` stored _and_ if gdpr is part of the config. 

Same for CCPA and USNat

PR #535, opened 2023-12-28
DIA-2934 show full text instead of truncated

Nevazhnovu:


andresilveirah:
@Nevazhnovu nice job with the commit messages. Since this PR has a couple different fixes, the messages helped me understand what was done.

PR #560, opened 2024-03-28
Make webConsents public

finnp:
Hey,

in our React Native app we would like to read out the webConsents ourselves, so we can inject them manually instead of relying on `WKWebView.loadConsent`.

For this we need to access the webConsents

Does this make sense?

Best,
Finn

### DIFF
--- a/ConsentViewController/Classes/Consents/SPUserData.swift
+++ b/ConsentViewController/Classes/Consents/SPUserData.swift
@@ -70,7 +70,7 @@ public class SPConsent<ConsentType: Codable & Equatable & NSCopying>: NSObject, 
     /// - SeeAlso: `SPUSNatConsent`
     public let usnat: SPConsent<SPUSNatConsent>?
 
-    var webConsents: SPWebConsents { SPWebConsents(
+    public var webConsents: SPWebConsents { SPWebConsents(
         gdpr: .init(uuid: gdpr?.consents?.uuid, webConsentPayload: gdpr?.consents?.webConsentPayload),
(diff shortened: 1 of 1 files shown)


andresilveirah:
@finnp @thekorn it does make sense. Just want to add a word of caution, the `webConsents` property is a "blackbox" to the SDK, in other words, we can't make any guarantee upon its contents.

If you're ok with that, I think it makes sense to make it public. 

Changing the topic a bit:
* @finnp we're planing to develop our own RN SDK and have it released this quarter. Would you be interested in give us feedback and share knowledge about it?
* @thekorn really cool stuff what you develop on https://github.com/thekorn/sourcepoint_unified_cmp . We have Flutter support in our Roadmap and I'd love to get some feedback as well.

thekorn:
@andresilveirah nice!

regarding the "blackbox" thingy: everything is better than just reverse engineering this blackbox in any integration of this package, I guess. and for me, this web consent is more or less just a "nice to have" - so having no guarantee is just fine for me

regarding the flutter plugin, and your roadmap @andresilveirah - sure, lets talk. im also open to just hand this integration over to you, the sourcepoint org, or keep it in mine, whatever you prefer.

finnp:
I also think that it's a good first step to make it public. 👍 

It's not part of the documentation so I would argue you make no guarantees for it. But for special cases like a React Native or Flutter integration it's very helpful.

I'm also happy to give feedback about a react native implementation. Great to see that you are planning on working on it.

andresilveirah:
Thank you for the feedback @finnp & @thekorn .

PR #561, opened 2024-04-02
DIA-2963 fix `onError` not called for timeouts

andresilveirah:
* add new Errors cases
* make sure network is interrupted on error
* refactor graceful degradation condition


andresilveirah:
@Nevazhnovu I added a few new error classes to `SPError` do you need to create the equivalent on the Unity SDK?

PR #566, opened 2024-04-18
[HCD 503] Check MSPS consent status

tom-win87:


tom-win87:
Please check code example

PR #570, opened 2024-05-06
[HCD 516] Add default language precedence

tom-win87:


andresilveirah:
@tom-win87 
> the user's browser language will always take precedence over the language set in the SDK
It's a bit stronger than that, if the `user browser default` is set, the language set in the SDK is ignored and the browser's language will be used instead, if that is not supported, the default language is used as a fallback.

What I mean to say is, `user browser default` not only takes precedence, but causes the rendering app to completely ignore SDK setting.

tom-win87:
@andresilveirah Thank you for the clarification. I've updated the note using the stronger language

PR #577, opened 2024-07-26
Add `rejectAll` to `SPSDK` protocol and implementation for that

Nevazhnovu:


andresilveirah:
@Nevazhnovu please check which callbacks are called after calling rejectAll new feature.

PR #598, opened 2025-03-07
DIA-5293 update `mobile-core`

andresilveirah:


Nevazhnovu (approved):
My only concern is 2 test suites not passing. Otherwise, it's looking good!

andresilveirah:
Thanks for the review @Nevazhnovu . I believe there's an issue with the test runner, let me re-run the tests.

PR #603, opened 2025-03-28
Fallback to default `fontSize` for tvOS

Nevazhnovu:
[DIA-5170](https://sourcepoint.atlassian.net/browse/DIA-5170) CMP Not Displaying on tvOS (Basque Country Public TV)


andresilveirah:
@Nevazhnovu let's not go ahead with this fallback. If the weren't by the SDK "breaking" because of this, we wouldn't have found the issue with the `fontSize` being a string (`"x rem"`) instead of a float.

PR #615, opened 2025-05-05
[DIA-5359] Add `preferences` to iOS SDK

Nevazhnovu:
[DIA-5359](https://sourcepoint.atlassian.net/browse/DIA-5359) [iOS] integrate mobile-core with preferences

### DIFF
--- a/ConsentViewController/Classes/SourcePointClient/Adapters.swift
+++ b/ConsentViewController/Classes/SourcePointClient/Adapters.swift
@@ -161,6 +161,8 @@ extension SPMobileCore.SPCampaignType {
             return .usnat
         case "IOS14":
             return .ios14
+        case "Preferences":
+            return .preferences
 
         default:
             return .unknown
@@ -453,6 +455,7 @@ extension SPCampaignType {
         case .ccpa: return .ccpa
         case .usnat: return .usnat
         case .ios14: return .ios14
+        case .preferences: return .preferences
         case .unknown: return .unknown
         }
     }
@@ -501,6 +504,7 @@ extension SourcepointClientCoordinator.State {
             ccpa: self.ccpa.toCore(metaData: self.ccpaMetaData),
             usNat: self.usnat.toCore(metaData: self.usNatMetaData),
             ios14: self.ios14.toCore(),
+            preferences: emptyPreferencesCampaign(),
             authId: self.storedAuthId,
             propertyId: Int32(propertyId),
             accountId: Int32(accountId),
@@ -593,6 +597,23 @@ extension SourcepointClientCoordinator.State.AttCampaign? {
     }
 }
 
+func emptyPreferencesCampaign() -> SPMobileCore.State.PreferencesState {
+    return SPMobileCore.State.PreferencesState(
+        metaData: SPMobileCore.State.PreferencesStatePreferencesMetaData(
+            configurationId: "",
+            additionsChangeDate: SPDate(date: Date.distantPast).toCore(),
+            legalDocLiveDate: nil
+        ),
+        consents: SPMobileCore.PreferencesConsent(
+            dateCreated: nil,
+            messageId: nil,
+            status: nil,
+            rejectedStatus: nil,
+            uuid: nil
+        )
+    )
+}
+
 extension SourcepointClientCoordinator.State.GDPRMetaData? {
     func toCore() -> SPMobileCore.State.GDPRStateGDPRMetaData {
         return SPMobileCore.State.GDPRStateGDPRMetaData.init(
@@ -662,7 +683,14 @@ extension SPCampaigns {
             gdpr: gdpr.toCore(),
             ccpa: ccpa.toCore(),
             usnat: usnat.toCore(),
-            ios14: ios14.toCore()
+            ios14: ios14.toCore(),
+            preferences: SPMobileCore.SPCampaign(
+                targetingParams: [:],
+                groupPmId: nil,
+                gppConfig: nil,
+                transitionCCPAAuth: nil,
+                supportLegacyUSPString: nil
+            )
         )
     }
 }
--- a/ConsentViewController/Classes/SPCampaignType.swift
+++ b/ConsentViewController/Classes/SPCampaignType.swift
@@ -8,7 +8,7 @@
 import Foundation
 
 @objc public enum SPCampaignType: Int, Equatable {
-    case gdpr, ios14, ccpa, usnat, unknown
+    case gdpr, ios14, ccpa, usnat, preferences, unknown
 }
 
 extension SPCampaignType: Codable {
@@ -20,6 +20,7 @@ extension SPCampaignType: Codable {
         case .ccpa: return "CCPA"
         case .ios14: return "ios14"
         case .usnat: return "usnat"
+        case .preferences: return "preferences"
         default: return "unknown"
         }
     }
@@ -30,6 +31,7 @@ extension SPCampaignType: Codable {
         case "CCPA": self = .ccpa
         case "ios14": self = .ios14
         case "usnat": self = .usnat
+        case "preferences": self = .preferences
         default: self = .unknown
         }
     }
--- a/ConsentViewController/Classes/SPCampaigns.swift
+++ b/ConsentViewController/Classes/SPCampaigns.swift
@@ -108,7 +108,7 @@ public typealias SPTargetingParams = [String: String]
 /// a active vendor list of that legislation.
 @objcMembers public class SPCampaigns: NSObject {
     public let environment: SPCampaignEnv
-    public let gdpr, ccpa, usnat, ios14: SPCampaign?
+    public let gdpr, ccpa, usnat, ios14, preferences: SPCampaign?
 
     override public var description: String {
         """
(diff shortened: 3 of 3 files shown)


ConsentViewController/Classes/Consents/SPPreferencesConsent.swift
```diff
@@ -0,0 +1,80 @@
...
+    public var uuid: String?
+    public var status, rejectedStatus: [PreferencesStatus]?
+
+
+    init(
+        dateCreated: SPDate,
+        messageId: String? = nil,
+        uuid: String? = nil,
+        status: [PreferencesStatus]?,
+        rejectedStatus: [PreferencesStatus]?
```
andresilveirah, line 23: Can we add `nil` as default params for both `status` and `rejectedStatus`?  and do they really need to be optional? I think it's a good practice to try and avoid optionals unless it's really a necessity.

ConsentViewController/Classes/Consents/SPPreferencesConsent.swift
```diff
@@ -0,0 +1,80 @@
...
+            - status: \(status ?? [])
+            - rejectedStatus: \(rejectedStatus ?? [])
+        )
+        """
+    }
+
+    public static func empty() -> SPPreferencesConsent { SPPreferencesConsent(
+        dateCreated: .now(),
+        status: [],
+        rejectedStatus: []
```
andresilveirah, line 49: once `status` and `rejectedStatus` have defaults, we won't need to declare them here.

ConsentViewController/Classes/Consents/SPPreferencesConsent.swift
```diff
@@ -0,0 +1,80 @@
...
+    )}
+
+    public func copy(with zone: NSZone? = nil) -> Any {
```
andresilveirah, line 52: Why is that `zone` parameter there?

ConsentViewController/Classes/Consents/SPPreferencesConsent.swift
```diff
@@ -0,0 +1,80 @@
...
+            status: status,
+            rejectedStatus: rejectedStatus
+        )
+    }
+}
+
+extension SPPreferencesConsent {
+    public struct PreferencesStatus: Codable {
+        var categoryId: Int
+        var channels: [PreferencesChannels]?
```
andresilveirah, line 66: Same here, could we make channels not optional and have a empty collection as default?

ConsentViewController/Classes/Consents/SPPreferencesConsent.swift
```diff
@@ -0,0 +1,80 @@
...
+        var changed: Bool?
+        var dateConsented: SPDate?
+        var subType: PreferencesSubType? = PreferencesSubType.Unknown
```
andresilveirah, line 69: and maybe add nil as default of `changed` and `dateConsented`

ConsentViewController/Classes/Consents/SPUserData.swift
```diff
@@ -66,10 +66,14 @@ public class SPConsent<ConsentType: Codable & Equatable & NSCopying>: NSObject,
     /// - SeeAlso: `SPCCPAConsent`
     public let ccpa: SPConsent<SPCCPAConsent>?
 
-    /// Consent data for USNat. This attribute will be nil if your setup doesn't include a CCPA campaign
+    /// Consent data for USNat. This attribute will be nil if your setup doesn't include a USNat campaign
```
andresilveirah, line 69: nice catch 👍

ConsentViewController/Classes/SourcePointClient/Adapters.swift
```diff
@@ -255,6 +257,48 @@ extension SPMobileCore.USNatConsent {
...
+        case .termsandconditions: return .TermsAndConditions
+        case .termsofsale: return .TermsOfSale
+        case .unknown: return .Unknown
+        default: return .Unknown
+        }
+    }
+}
+
+extension SPMobileCore.PreferencesConsent.PreferencesStatusPreferencesChannels {
+    func toNative() -> SPPreferencesConsent.PreferencesChannels { return .init(channelId: Int(self.channelId), status: self.status) }
```
andresilveirah, line 275: when a function has a single expression, the `return` keyword can be omitted. 

Could you remove it?

ConsentViewController/Classes/SourcePointClient/Adapters.swift
```diff
@@ -255,6 +257,48 @@ extension SPMobileCore.USNatConsent {
...
+}
+
+extension SPMobileCore.PreferencesConsent.PreferencesStatus {
+    func toNative() -> SPPreferencesConsent.PreferencesStatus {
+        return .init(
```
andresilveirah, line 280: same goes here, removing `return`

ConsentViewController/Classes/SourcePointClient/Adapters.swift
```diff
@@ -255,6 +257,48 @@ extension SPMobileCore.USNatConsent {
...
+            changed: self.changed?.boolValue,
+            dateConsented: SPDate(string: self.dateConsented?.instantToString() ?? SPDate.now().originalDateString),
+            subType: self.subType.toNative()
+        )
+    }
+}
+
+extension SPMobileCore.PreferencesConsent {
+    func toNative() -> SPPreferencesConsent {
+        return .init(
```
andresilveirah, line 292: and here also

ConsentViewController/Classes/SourcePointClient/Adapters.swift
```diff
@@ -255,6 +257,48 @@ extension SPMobileCore.USNatConsent {
...
+            dateCreated: SPDate(string: self.dateCreated?.instantToString() ?? SPDate.now().originalDateString),
+            messageId: String(Int(truncating: self.messageId ?? 0)),
```
andresilveirah, line 294: are you sure `messageId` shouldn't be optional? it feels strange having to pass a default value of `0`  to it. I'm concerned this might affect reporting.

ConsentViewController/Classes/SourcePointClient/Adapters.swift
```diff
@@ -282,12 +326,22 @@ extension SPUserDataSPConsent<USNatConsent>? {
     }
 }
 
+extension SPUserDataSPConsent<PreferencesConsent>? {
+    func toNative() -> SPConsent<SPPreferencesConsent>? {
+        return SPConsent<SPPreferencesConsent>.init(
```
andresilveirah, line 331: and again here regarding remove `return`

ConsentViewController/Classes/SourcePointClient/Adapters.swift
```diff
@@ -593,6 +649,23 @@ extension SourcepointClientCoordinator.State.AttCampaign? {
     }
 }
 
+func emptyPreferencesCampaign() -> SPMobileCore.State.PreferencesState {
```
andresilveirah, line 652: What do you think of renaming this method to `emptyPreferencesState`?

ConsentViewController/Classes/SourcePointClient/Adapters.swift
```diff
@@ -501,6 +556,7 @@ extension SourcepointClientCoordinator.State {
             ccpa: self.ccpa.toCore(metaData: self.ccpaMetaData),
             usNat: self.usnat.toCore(metaData: self.usNatMetaData),
             ios14: self.ios14.toCore(),
+            preferences: emptyPreferencesCampaign(),
```
andresilveirah, line 559: Why are we initializing state with empty preferences data?

ConsentViewController/Classes/Consents/SPPreferencesConsent.swift
Nevazhnovu: @andresilveirah This is done to conform to the obj-c protocol `NSCopying`, the same way as other campaigns do

ConsentViewController/Classes/SourcePointClient/Adapters.swift
Nevazhnovu: I was referencing https://github.com/SourcePointUSA/android-cmp-app/pull/859 PR

ConsentViewController/Classes/SourcePointClient/Adapters.swift
Nevazhnovu: @andresilveirah `messageId` in Swift is a `String?` and  `messageId` in Kotlin is an `Int?`. 
Casting Kotlin's optional `Int?` into Swift's `Int` is tricky and requires a `?? 0` fallback, as Swift's `Int` cast does not want null.
The best we can do here is
> messageId: self.messageId != nil ? String(Int(truncating: self.messageId ?? 0)) : nil,

ConsentViewController/Classes/SourcePointClient/Adapters.swift
Nevazhnovu: @andresilveirah this is used as a part of data migration from native SDK data storage to mobile core data storage on the first-ever launch. 
There are 2 use cases:
1. The user had some other campaign data in the native SDK storage. Since we previously didn't have `preferences`, we need to fill the state with an empty `preferences` campaign object.
2. If there is no consent exists in data storage, neither in native nor in core, we need an empty `preferences` to initialize the core's data storage

ConsentViewController/Classes/Consents/SPPreferencesConsent.swift
Nevazhnovu: Linters say to remove nil. 
>Redundant Optional Initialization Violation: Initializing an optional variable with nil is redundant (redundant_optional_initialization)

PR #635, opened 2025-09-03
Fix compile issue

jonas-collide:
Seems like the iOS SDK does not compile since 7.12.2


andresilveirah:
Thanks for the PR @jonas-collide . Can you help me understand your local setup? The SDK compiles just fine for me, using:
```
Apple Swift version 6.1.2 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)
Target: arm64-apple-macosx15.0
```

PR #650, opened 2026-05-18
avoid calling `messageUIDelegate.loaded` repeatedly

andresilveirah:
If the rendering app dispatches the event `sp.showMessage` multiple times and the consentUI is being presented, clients' app can crash if their code is not guarding against re-presenting the consent ViewController.

Almost entirely inspired on #649

### DIFF
--- a/ConsentViewController/Classes/Views/iOS/SPWebMessageViewController.swift
+++ b/ConsentViewController/Classes/Views/iOS/SPWebMessageViewController.swift
@@ -289,7 +289,9 @@ import WebKit
     func onMessageReady() {
         timeoutWorkItem.cancel()
         webview?.evaluateJavaScript("window.spLegislation=\"\(self.campaignType.rawValue)\"")
-        messageUIDelegate?.loaded(self)
+        if !isBeingPresented, view.window == nil { // make sure current ViewController is not being presented
+            messageUIDelegate?.loaded(self)
+        }
     }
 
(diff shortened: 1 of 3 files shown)


sampaioroberto:
@andresilveirah  Thanks again for merging and crediting #649! 😊 

One small concern with the guard: `isBeingPresented` and `view.window` only change after the host app calls `present()`, which happens inside the `DispatchQueue.main.async` in `onSPUIReady`. A second event arriving before that block runs would still pass the check and call `loaded()` twice.

There's also a minor issue in the test: `loadMessage()` instantiates `GenericWebMessageViewController`, not `MockedGenericWebMessageViewController`, so the `mockIsBeingPresented` override is never active in the duplicated-event scenario. The test passes because `toEventually` catches the count between the two deliveries rather than the guard actually blocking the second call.

Happy to be wrong if I'm missing something though!

andresilveirah:
Hey @sampaioroberto you are absolutely right about both, the test and the racing condition when calling  `loaded()`. In the end, I think it'll be easier to deal with it with the flag as you first suggested.

sampaioroberto:
Hey @andresilveirah Glad it helped! Are you planning to open a new PR with the flag approach, or would you like me to reopen #649?

andresilveirah:
@sampaioroberto this fix should be in version `7.12.10` thanks again for reporting and putting a fix to it.

PR #651, opened 2026-05-20
use flag to control if `MuessageUIDelegate.loaded()` was called

andresilveirah:
Guard against calling `MuessageUIDelegate.loaded()` multiple times if the `sp.showMessage` event is dispatched repeatedly. 

Reference #650 #649


andresilveirah:
@sampaioroberto FYI

sampaioroberto (approved):
Nice, the `toNever` is way better than what I had. And the reset in `viewWillDisappear` is a good catch. LGTM!

